PR #67, opened 2026-05-14
fix(frontend): match conversations by id||uuid in selection and lifecycle reducers (EVO-1145)

nickoliveira23:
## Summary

Follow-up to EVO-1118 / EVO-1062. WebSocket frames can arrive keyed either by `Conversation.id` or `Conversation.uuid`; comparing only against `id` silently misses uuid-keyed payloads and is the root cause of the duplicated rows the C3 review flagged on PR #58 (`evo-ai-frontend-community`).

- New `matchesConversationId` helper in `src/utils/chat/conversationMatcher.ts` (matches by `id || uuid`, guards null conv + empty `idStr`).
- `ConversationsContext.findConversationByAnyId` delegates to the helper so the reducer and the context lookup share a single contract.
- `UPDATE_CONVERSATION` resolves the canonical conversation first, then drives `isSelected`, `syncedUnreadCount`, and `updatedSelectedData` off it. The unread sync now consults BOTH the canonical-id key and the selection key, because `selectConversation` canonicalizes to `uuid` while `SET_CONVERSATIONS` seeds `unreadCounts` by `String(conv.id)`.
- `ADD_CONVERSATION`, `REMOVE_CONVERSATION`, and `UPDATE_UNREAD_COUNT` apply the same matcher; `REMOVE_CONVERSATION` also resolves the canonical `unreadCounts` key before pruning.
- `UPDATE_UNREAD_COUNT` now clears the persisted "marked as read" entry in `localStorage` when the count rises above zero — fixes the bug where mark-as-unread silently lost the badge after F5 (`SET_CONVERSATIONS` was honoring a stale `readConversations[id] = true`).

## Changed Files

- `src/utils/chat/conversationMatcher.ts` (new, 16 lines)
- `src/contexts/chat/ConversationsContext.tsx`
- `src/contexts/chat/ConversationsContextReducer.ts`
- `src/contexts/chat/ConversationsContext.spec.tsx` (new spec, 18 tests)

## Validation

- `pnpm exec tsc -b --noEmit` — clean.
- `pnpm lint` — no new errors on changed files (the existing repo-wide lint debt is out of scope; the only warning on `ConversationsContext.tsx` is pre-existing at line 816 and untouched by this patch).
- `pnpm exec vitest run src/contexts/chat/ConversationsContext.spec.tsx` — **18/18 passing**. Targeted run because the full vitest suite hangs locally (separate issue under triage).
- Manual smoke validated: open conv → badge clears; mark as unread → F5 → badge persists; resolve/snooze does not duplicate the row; selection by uuid keeps right panel in sync.

## Out of Scope (carry-forward)

The following lifecycle cases still match by `String(conv.id)` only and should follow the same normalization in a future card: `UPDATE_CONVERSATION_LAST_ACTIVITY`, `INCREMENT_UNREAD_COUNT`, `ADD_HIDDEN_CONVERSATION`, `REVEAL_HIDDEN_CONVERSATIONS`, `UPDATE_CONTACT_IN_CONVERSATIONS`, `CLEANUP_DUPLICATES`. The pre-existing `SELECT_CONVERSATION` reducer also fails to clear `conv.unread_count` on the list row when SELECT is canonicalized to uuid; downstream `UPDATE_CONVERSATION` frames now compensate, but the SELECT reducer itself should also be normalized in the follow-up.

## Linked Issue

- [EVO-1145](https://linear.app/evoai/issue/EVO-1145/evo-1118-follow-up-corrigir-iduuid-match-no-update-conversation)

### DIFF
--- a/src/contexts/chat/ConversationsContext.spec.tsx
+++ b/src/contexts/chat/ConversationsContext.spec.tsx
@@ -1,15 +1,29 @@
 import { beforeEach, describe, expect, it } from 'vitest';
 import { conversationsReducer } from './ConversationsContextReducer';
 import { initialState } from '@/types/chat/conversations';
+import type { Conversation } from '@/types/chat/api';
+import { matchesConversationId } from '@/utils/chat/conversationMatcher';
 
-const makeConversation = (overrides: Record<string, unknown> = {}) =>
-  ({
+// Test fixtures admit numeric ids because the backend / WebSocket can deliver
+// either form despite the strict `Conversation.id: string` typing. Cast back
+// to `Conversation` at the boundary so the reducer sees its expected shape.
+type ConversationFixture = Omit<Partial<Conversation>, 'id' | 'uuid'> & {
+  id: string | number;
+  uuid?: string | null;
+};
+
+const makeConversation = (
+  overrides: Partial<ConversationFixture> = {},
+): Conversation => {
+  const base: ConversationFixture = {
     id: 'conversation-1',
     uuid: 'conversation-1',
     unread_count: 5,
     status: 'open',
     ...overrides,
-  }) as any;
+  };
+  return base as unknown as Conversation;
+};
 
 describe('ConversationsContext reducer', () => {
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


dpaes (approved):
LGTM — approving. ACs all met, 18/18 spec tests pass locally, typecheck clean on changed files (pre-existing `Chat.tsx` error is on `develop`, unrelated). C3 root-cause fix is solid and the helper extraction lands the contract in a single place.

Detailed review in PT-BR posted on [EVO-1145](https://linear.app/evoai/issue/EVO-1145).

**Worth a follow-up card** (no need to block this PR): in `UPDATE_CONVERSATION`, when match happens via `uuid`, the spread `{ ...conv, ...action.payload }` rewrites `conv.id` to the payload's uuid string. Today this is contained because `selectConversation` canonicalizes selection to uuid and most downstream dispatchers go through `findConversationByAnyId` — but the carry-forward reducers (`INCREMENT_UNREAD_COUNT`, `UPDATE_CONVERSATION_LAST_ACTIVITY`, `SELECT_CONVERSATION` row-clear at line 181, `CLEANUP_DUPLICATES`, etc.) still match by `String(conv.id)` only and will silently miss the row if any caller still passes the numeric id. Cheapest fix is preserving identity at merge time:

```ts
matchesConversationId(conv, conversationId)
  ? { ...conv, ...action.payload, id: conv.id, uuid: conv.uuid ?? action.payload.uuid, unread_count: syncedUnreadCount }
  : conv,
```

Also worth pairing: `SELECT_CONVERSATION` line 181 still uses `conv.id === action.payload`, leaving the list-row badge stale until the next `UPDATE_CONVERSATION` compensates — one-line switch to `matchesConversationId(conv, payloadStr)` closes it.

PR #98, opened 2026-05-20
fix(frontend): unwrap envelope + defensive guards + ErrorBoundary in SessionsViewer (EVO-1254)

nickoliveira23:
## Summary

Fixes Pain #1 from the Flow Builder discovery (EVO-1237): the Sessions Viewer crashed with `TypeError: Cannot read properties of undefined (reading 'active')` when opened from the Journey Editor, leaving the page in a black-screen state.

Root cause was **not** the auth regression the card hypothesised. The evo-flow backend wraps every response in `{ success: true, data: ... }` via a global `ResponseTransformInterceptor` (`evo-flow/src/main.ts:134`). The core `journeyService` methods (`getJourneys`, `getJourney`) use the `extractData` helper to unwrap that envelope — but the session-related methods returned `response.data` raw, so `stats` ended up shaped as `{ success: true, data: { byStatus: {...} } }` instead of `{ byStatus: {...} }`. Line 213's access on `stats.byStatus.active` was reading `undefined.active`.

## What changed

- **Envelope unwrap** in `journeyService.getJourneySessions`, `getJourneySessionStats`, `getJourneySession`, `cancelJourneySession`, `bulkDeleteJourneySessions`. `deleteJourneySession` is unchanged (returns void).
- **Defensive JSX guards** on `SessionsViewer`: every `stats.byStatus.X` is now `stats.byStatus?.X ?? 0`; `stats.total` is `stats.total ?? 0`. Stats grid renders zeros instead of crashing when the payload is incomplete.
- **Typed stats state**: `stats: any` is replaced with a `SessionStats` interface (`total?: number; byStatus?: Partial<Record<SessionStatus, number>>`). `SessionStatus` is now a single type alias reused by both `JourneySession['status']` and `SessionStats['byStatus']`.
- **Explicit return type** on `getJourneySessionStats` mirrors the new `SessionStats` shape — self-documenting and forces callers to handle missing fields.
- **`<ErrorBoundary>` wrap** around the entire `JourneyFlowEditor` page using the shared component at `src/components/ErrorBoundary.tsx` (same one used by Chat). Internal `JourneyFlowEditor` stays as-is; a new outer `JourneyFlowEditorPage` is the default export and renders the boundary. Any uncaught error inside the canvas, palette, header, or node modals now shows the i18n fallback (`common.errorBoundary.*`) with a Retry button instead of a black screen.
- **`data-testid`** on every stats card (`sessions-stats-grid`, `sessions-stat-{total|active|waiting|completed|failed|cancelled}`) for stable selectors in tests and future automation.

## Security

- No new endpoints, no new authN/authZ surfaces, no new query patterns. Pure UI defensive layer + response shape handling.
- ErrorBoundary swallows runtime errors but does NOT swallow `console.error`, so dev telemetry remains intact.

## Test plan

- [x] `pnpm exec tsc -b --noEmit` — clean
- [x] `pnpm exec eslint src/components/journey/SessionsViewer.tsx src/services/journeys/ src/pages/Customer/Journey/JourneyFlowEditor.tsx` — no new errors; the 12 pre-existing `any` errors in `journeyService.ts` and `JourneyFlowEditor.tsx` are unchanged from `develop`.
- [x] `pnpm test src/services/journeys/journeyService src/components/journey/SessionsViewer` — **12/12 passing** (6 service + 6 component).
- [x] Manual smoke (user-confirmed): open journey → click "Ver Sessões" → modal renders with all 6 stats cards at 0, "No sessions found" empty state, no crash, no console errors.
- [x] AC #1 (no 401, request returns 200 with correct shape) — covered by service-level envelope unwrap test.
- [x] AC #2 (payload inesperado não crasha) — 5 spec cases cover `stats = {}`, `stats = { total: N }`, partial `byStatus`, rejected promise, and the **legacy envelope leak** (regression test that locks in defence-in-depth even if the service-side unwrap is undone in the future).
- [x] AC #3 (error boundary around Flow Builder) — verified by import + wrap; visual fallback inherited from the shared `<ErrorBoundary>`.
- [x] AC #4 (3 specs pass without flaky behaviour) — 12 cases, all hermetic (no real timers, no real network).

## Changed Files

- `src/services/journeys/journeyService.ts` — envelope unwrap on 5 session methods + typed return on stats
- `src/components/journey/SessionsViewer.tsx` — SessionStats interface, SessionStatus DRY type, defensive guards, data-testid attributes
- `src/components/journey/SessionsViewer.spec.tsx` — new, 6 component tests
- `src/services/journeys/journeyService.spec.ts` — new, 6 service tests
- `src/pages/Customer/Journey/JourneyFlowEditor.tsx` — ErrorBoundary wrap (renamed default export to `JourneyFlowEditorPage`, internal `JourneyFlowEditor` unchanged)

## QA Notes

Three adversarial code-review passes (`/evo-code-review`) were applied during development. The third pass surfaced a regression-resistance gap and tightening opportunities — all addressed inline before requesting review:

- **First pass:** envelope handling was the real bug (not auth as the card description suspected). Plan adjusted to swap auth investigation for envelope unwrap.
- **Second pass:** all 8 first-pass findings shipped in the implementation commits.
- **Third pass (6 findings):** added an envelope-leak regression test, tightened typing on `getJourneySessionStats`, DRY'd the SessionStatus type, swapped brittle text-based test selectors for `data-testid`-scoped queries. One observation (paused stats card missing from grid) is **pre-existing in develop** and filed separately as EVO-1417 so this PR stays focused.

## Out of scope (deferred)

- **Auth regression investigation** — the card hypothesised a client-side auth bug; investigation showed the same axios instance is used by working journey methods. The "401" symptom in the discovery doc was likely a transient session-expiry mixed with the envelope crash. If a real 401 reappears on this route, it would be a separate auth-service issue.
- **Sentry/logger integration** — the card mentions "envia erro para logger configurado"; the project has no Sentry config (no `@sentry/*` imports anywhere). ErrorBoundary keeps the existing `console.error` behaviour. If telemetry is added later, it lives in `ErrorBoundary.tsx`.
- **Paused stats card missing** — filed as [EVO-1417](https://linear.app/evoai/issue/EVO-1417/sessionsviewer-stats-grid-is-missing-the-paused-counter-card), separate UI card.

## Linked Issue

- https://linear.app/evoai/issue/EVO-1254/101-corrigir-401-no-sessions-viewer-crash-js-error-boundary

## Summary by Sourcery

Handle wrapped session stats responses and harden the Sessions Viewer and Flow Editor against runtime crashes.

Bug Fixes:
- Unwrap the evo-flow response envelope for journey session APIs so the Sessions Viewer receives correctly shaped stats data.
- Prevent Sessions Viewer from crashing when session stats are missing, partially populated, or returned in an unexpected envelope shape.

Enhancements:
- Type the session stats and status model shared between services and SessionsViewer to better reflect optional fields and reduce reliance on any.
- Wrap the Journey Flow Editor page in a shared ErrorBoundary so unexpected errors render a recoverable fallback instead of a blank screen.
- Add data-testid attributes to session stats cards to support stable, targeted testing of the stats grid.

Tests:
- Add component tests for SessionsViewer covering defensive handling of empty, partial, and malformed stats responses, as well as stats request failures.
- Add service-level tests around journey session APIs to validate correct unwrapping and shaping of stats responses.

### DIFF
--- a/src/components/journey/SessionsViewer.spec.tsx
+++ b/src/components/journey/SessionsViewer.spec.tsx
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { render, screen, waitFor } from '@testing-library/react';
+import { render, screen, waitFor, within } from '@testing-library/react';
 import { SessionsViewer } from './SessionsViewer';
 
 vi.mock('@/services', () => ({
@@ -29,6 +29,11 @@ const baseProps = {
   onClose: () => undefined,
 };
 
+function statCount(testId: string) {
+  const card = screen.getByTestId(testId);
+  return within(card).getByText(/^\d+$/).textContent;
+}
+
 beforeEach(() => {
   vi.clearAllMocks();
 });
@@ -52,11 +57,14 @@ describe('SessionsViewer — defensive stats handling', () => {
     render(<SessionsViewer {...baseProps} />);
 
     await waitFor(() => {
-      expect(screen.getByText('12')).toBeTruthy();
+      expect(screen.getByTestId('sessions-stats-grid')).toBeTruthy();
     });
-    expect(screen.getByText('3')).toBeTruthy();
-    expect(screen.getByText('2')).toBeTruthy();
-    expect(screen.getByText('4')).toBeTruthy();
+    expect(statCount('sessions-stat-total')).toBe('12');
+    expect(statCount('sessions-stat-active')).toBe('3');
+    expect(statCount('sessions-stat-waiting')).toBe('2');
+    expect(statCount('sessions-stat-completed')).toBe('4');
+    expect(statCount('sessions-stat-failed')).toBe('1');
+    expect(statCount('sessions-stat-cancelled')).toBe('1');
   });
 
   it('does NOT crash when stats has no byStatus field and renders zeros instead', async () => {
@@ -70,11 +78,11 @@ describe('SessionsViewer — defensive stats handling', () => {
     expect(() => render(<SessionsViewer {...baseProps} />)).not.toThrow();
 
     await waitFor(() => {
-      const zeros = screen.getAllByText('0');
-      // total + 6 status cards (active, waiting, paused doesn't render in JSX
-      // but completed, failed, cancelled do — so 5 zero cards on top of `total`)
-      expect(zeros.length).toBeGreaterThanOrEqual(5);
+      expect(screen.getByTestId('sessions-stats-grid')).toBeTruthy();
     });
+    expect(statCount('sessions-stat-total')).toBe('0');
+    expect(statCount('sessions-stat-active')).toBe('0');
+    expect(statCount('sessions-stat-cancelled')).toBe('0');
   });
 
   it('does NOT crash when stats is an empty object and renders zeros', async () => {
@@ -88,8 +96,10 @@ describe('SessionsViewer — defensive stats handling', () => {
     expect(() => render(<SessionsViewer {...baseProps} />)).not.toThrow();
 
     await waitFor(() => {
-      expect(screen.getAllByText('0').length).toBeGreaterThanOrEqual(5);
+      expect(screen.getByTestId('sessions-stats-grid')).toBeTruthy();
     });
+    expect(statCount('sessions-stat-total')).toBe('0');
+    expect(statCount('sessions-stat-failed')).toBe('0');
   });
 
   it('does NOT crash when byStatus is partial (missing some statuses) — missing ones render as 0', async () => {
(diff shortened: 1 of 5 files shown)


dpaes (approved):
## Code Review — EVO-1254 — ✅ Approved (with minor follow-ups)

Reviewed against the actual code at `feat/EVO-1254` (not just the diff): `ErrorBoundary.tsx`, `apiHelpers.extractData`, `apiEvoFlow.ts`, the full `SessionsViewer.tsx`, routing, and the locale files.

**Verdict:** Approving. No functional bug, no security concern. Root cause was correctly re-diagnosed (response envelope, not auth). One explicitly-required spec is missing (M1) plus a few low-severity items — all acceptable as follow-ups.

### Verified
- **Root cause / AC#2:** `extractData` (`src/utils/apiHelpers.ts`) unwraps `response.data.data` and falls back to raw — the documented behavior the regression test relies on is real. The 5 session methods now match the working core methods. JSX guards (`stats.total ?? 0`, `stats.byStatus?.X ?? 0`) hold the line even if the service-side unwrap is undone. Solid.
- **AC#3 (ErrorBoundary):** Placement is correct — `routes/index.tsx` imports the **default** export, which is now `JourneyFlowEditorPage` → `<ErrorBoundary>` → editor (the `index.ts` barrel also forwards the wrapped version). Fallback + retry button live in the shared component; `base.errorBoundary.{title,description,retry}` exist in pt-BR/en (and es/it/fr).
- **AC#4:** 12 hermetic tests with real assertions (mocked service/i18n/toast, no timers, no network).

### 🟡 MEDIUM
**M1 — The error-boundary spec required by the card is missing.** "Escopo — dentro" lists three specs: (a) auth header, (b) defensive, **(c) error boundary captures the error and renders the fallback**. There is **no ErrorBoundary spec anywhere in the repo** — neither for the shared component nor for the JourneyFlowEditor wrap; the PR treats it as "verified by import + wrap". Cheapest gap to close: a ~5-line test rendering `<ErrorBoundary>` around a throwing child and asserting the fallback + "Try again" button. (Spec (a) "auth header" is also absent, replaced by the envelope tests — defensible since auth was descoped.)

### 🟢 LOW
- **L1 — Logger requirement partially met.** AC#3 asks the boundary to "send the error to the configured logger"; it uses `console.error` in `componentDidCatch`. Acceptable given there is no Sentry/telemetry in the project — flag for the PO / future telemetry follow-up.
- **L2 — "Try again" doesn't reload.** `handleRetry` only resets `hasError`; on a deterministic render error it re-renders the same tree and lands back on the fallback. Shared-component behavior (applies to Chat too), not introduced here — noting only.
- **L3 — Latent crash still in the sessions LIST render (unchanged region of a touched file).** `const config = statusConfig[session.status]` then `config.borderColor` with no optional chaining, while `getStatusIcon` uses `config?.`. An unexpected backend status still throws here (now caught by the new boundary instead of black-screening). Since the PR's goal includes hardening SessionsViewer, `config?.borderColor` would close the inconsistency.

### ⚪ Nits
- **N1:** PR body says the fallback uses `common.errorBoundary.*`; the real key path is `common:base.errorBoundary.*`. Docs only.
- **N2:** `extractData` unwraps any `response.data.data` — latent footgun if a payload legitimately has a top-level `data`. Not triggered by current session shapes.
- **N3:** Only `getJourneySessionStats` got a tightened return type; the other 4 stay loose — explicitly scoped out, fine.

### On AC#1 ("fix the 401 / returns 200")
The *outcome* is demonstrated: the user-confirmed smoke opened Sessions with no 401 and no crash, and since `apiEvoFlow`'s response interceptor calls `clearUser()` on any 401, a real 401 would have logged the user out (very visible). What's descoped is the *root-cause hypothesis* (auth), not the result. Low-priority suggestion: have Davidson re-test on his original repro environment to confirm the 401 doesn't recur there.

**Strengths:** correct root-cause diagnosis overturning the card's hypothesis; unwrap aligned with the existing `extractData` pattern; regression test locking the contract; typed `SessionStats` / DRY `SessionStatus`; `data-testid` for stable selectors; ErrorBoundary reuse (no new deps or i18n keys); clean commits, File List matches the diff.

---
🤖 Reviewed via `/evo-code-review`

PR #102, opened 2026-05-21
feat(frontend): refactor flow editor autosave to dedicated store + state machine + IndexedDB recovery (EVO-1258)

nickoliveira23:
## Summary

Pain #9 of the discovery 8.1 ([EVO-1258](https://linear.app/evoai/issue/EVO-1258), [10.22]) called for a re-architecture of the Flow Builder autosave / dirty-state subsystem. Pre-refactor, three parallel `setInterval` autosave timers ran on every dirty cycle (one in `BaseFlowEditor`, two in `JourneyFlowEditor`), two parallel `hasUnsavedChanges` `useState` instances flipped independently, and there was no persistence layer to recover unsaved work.

This PR introduces `useFlowEditorStore` (Zustand) as the single source of truth, with a 4-state machine (`idle | dirty | saving | error`), a real debounced autosave (5s, resets on every edit), an IndexedDB snapshot for recovery on reopen, and a per-device `localStorage` mark for the "last saved" timestamp.

### Architecture

| File | Purpose |
|---|---|
| `src/store/flowEditor/useFlowEditorStore.ts` | Zustand store with the state machine, autosave/IDB/retry timers, atomic-update guard. |
| `src/store/flowEditor/idbSnapshot.ts` | Thin `idb-keyval` wrapper for per-journey snapshots. JSON-clones the payload to strip xyflow callbacks (DataCloneError defence). |
| `src/store/flowEditor/lastSavedMark.ts` | `localStorage` mark for the "last saved at" timestamp. Preferred over `response.updatedAt` because the server timestamp has been observed coming back stale right after a save (see "Known limitations" below). |
| `src/store/flowEditor/README.md` | API + state machine diagram + recovery flow + singleton constraint + anti-patterns. |

### Acceptance Criteria — all five covered

- **AC #1 (recovery prompt):** on next mount, if the IDB snapshot is newer than `lastSavedAt` AND diverges from server data, the consumer renders an `<AlertDialog>` with "Recuperar versão de HH:MM?" — Recuperar loads into `currentSnapshot` and re-arms autosave; Descartar clears IDB.
- **AC #2 (5s autosave with state transitions):** `setFlow` / `setVariables` flip to `dirty` and arm a debounced 5s timer that resets on every subsequent edit. When the timer fires, the registered consumer trigger is called → `beginSave()` → API → `commitSave(savedAt, syncedSnapshot)` or `failSave(message)`.
- **AC #3 (manual save cancels timer):** `beginSave` clears the pending autosave timer; a manual click during a pending tick saves immediately and does not duplicate.
- **AC #4 (failure + 30s retry):** `failSave` flips to `error`, surfaces the message in a `<FlowFeedbackBanner variant="error">`, and arms a single auto-retry in 30s. The banner text changes between "Retrying in 30s" and "Click Save to try again" based on the `retryScheduled` flag (no false promises when no trigger is registered).
- **AC #5 (browser close confirmation):** a `beforeunload` listener registers while status is non-idle, so the browser shows its native unsaved-changes prompt.

### Atomic update

The card requested explicitly: *"Atomic update: mudanças no canvas bloqueadas durante save."*

This PR implements it by capturing the snapshot at `beginSave()` time and passing it to `commitSave(savedAt, syncedSnapshot)`. If the user edited mid-roundtrip (`currentSnapshot ≠ syncedSnapshot`), the store stays `dirty` and `serverSnapshot` is set to `syncedSnapshot` (NOT `currentSnapshot`) — we never lie about having synced edits the server never received. The mid-save edit re-arms autosave for the next cycle. Test added covering the canonical scenario.

### Adversarial review fixes

Self-review via `/evo-code-review` after the initial implementation surfaced 12 issues. All addressed in commit `b2df832`. Notable:

- **H1:** save was not atomic (described above).
- **H2:** retry timer leaked when the consumer's `useEffect` re-registered the trigger between `failSave` and the 30s tick. `unregister()` now clears the retry timer too.
- **M3:** module-level timer singletons documented as a single-mounted-editor contract (in code and README).
- **M5:** banner stopped promising "Retrying in 30s" when no trigger was registered. New `retryScheduled` boolean controls the banner copy.
- **L11:** `saveChanges` was being recreated on every `setJourney`, causing the autosave trigger to re-register (this was the vector for H2). Now `journey` lives in a `useRef`; `saveChanges` deps shrink to `[t]`.

## Security

Frontend-only refactor. No new endpoints, no auth surfaces, no validation changes. IndexedDB and localStorage are sandboxed per origin; keys include `journeyId` so cross-tenant leakage is not possible. The save error banner renders the `lastError` string as JSX text (React escapes), no XSS.

## Test plan

- `evo-ai-frontend-community: pnpm exec tsc -p tsconfig.app.json --noEmit` — clean
- `evo-ai-frontend-community: pnpm test src/store/flowEditor src/lib/useRelativeTime` — 48/48 passing
  - `useFlowEditorStore.spec.ts` — 26 tests (hydration, recovery selection, dirty propagation, debounce reset, save lifecycle, atomic update on mid-save edit, error → dirty on new edit, 30s auto-retry, all 5 retry cancellation paths, retryScheduled flag truthfulness, acceptRecovery defensive cleanup)
  - `idbSnapshot.spec.ts` — 8 tests (round-trip, journey isolation, overwrite, clear, 7-day staleness, function-stripping for DataCloneError defence)
  - `lastSavedMark.spec.ts` — 7 tests (round-trip, journey isolation, overwrite, clear, 30-day staleness, corrupted-value graceful fallback)
  - `useRelativeTime.spec.tsx` — 7 tests (initial, ticking cadences, cleanup on unmount, snap-to-wall-clock on date change for the EVO-1258 smoke regression)
- `evo-ai-frontend-community: pnpm exec eslint src/store/flowEditor src/pages/Customer/Journey/JourneyFlowEditor.tsx` — 7 problems, IDENTICAL to `origin/develop` baseline (all pre-existing `any` types in `renderConfigPanel` and `react-hooks/exhaustive-deps` on `loadJourney`/`saveChanges` deps of `t`)

Manual smoke covered: 5s autosave, manual save during pending tick (no duplicate), error → retry → success, refresh / close tab while dirty, recovery prompt on reopen after kill-tab, mid-save edit stays dirty after success.

## Changed Files

- `src/store/flowEditor/useFlowEditorStore.ts` (new, 380 lines)
- `src/store/flowEditor/useFlowEditorStore.spec.ts` (new, 26 tests)
- `src/store/flowEditor/idbSnapshot.ts` (new)
- `src/store/flowEditor/idbSnapshot.spec.ts` (new, 8 tests)
- `src/store/flowEditor/lastSavedMark.ts` (new)
- `src/store/flowEditor/lastSavedMark.spec.ts` (new, 7 tests)
- `src/store/flowEditor/README.md` (new)
- `src/pages/Customer/Journey/JourneyFlowEditor.tsx` — wired to store, removed 3 timers, added recovery prompt and error banner, added `friendlySaveErrorMessage` helper
- `src/lib/useRelativeTime.ts` — fix: snap `now` when input date changes (regression caught during smoke)
- `src/lib/useRelativeTime.spec.tsx` — test for the snap behaviour
- `src/i18n/locales/{en,es,fr,it,pt,pt-BR}/journey.json` — added `recoveryTitle`, `recoveryBody`, `recover`, `discard`, `saveErrorBanner`, `saveErrorBannerNoRetry`, `saveErrorMessages.{network,server,unauthorized,validation}` (10 new keys per locale)
- `package.json` + `pnpm-lock.yaml` — added `idb-keyval` (runtime) + `fake-indexeddb` (dev)

Net diff: **+1875 / −257 lines**.

## Known limitations / out of scope

- **`response.updatedAt` workaround:** the server timestamp from `journeyService.getJourney(id).updatedAt` has been observed coming back stale (header showed "Saved 2 hours ago" right after a fresh save). The frontend workaround in this PR (`lastSavedMark` localStorage) trusts our wall-clock at commit time. The proper fix — making the backend reliably bump `updated_at` on every save and serialise it correctly — should be filed as a separate Rails ticket. Not addressed here.
- **`BaseFlowEditor` internal duplication:** `JourneyFlowEditor` now passes `autoSave={false}` so `BaseFlowEditor`'s `useState hasUnsavedChanges` + `useEffect setInterval` is dead state when consumed from journey. Left in place because `BaseFlowEditor` is its only consumer today; a full cleanup of the base component is a separate refactor.
- **`pendingLeaveTarget` still in component state:** could move to the store, but crosses the boundary with `react-router`'s `useNavigate`. Deferred.
- **No mount-level integration test for the navigate/AlertDialog/status combo:** the store and helpers have 48 unit tests; a full `@testing-library/react` mount of `JourneyFlowEditor` requires mocking xyflow + `journeyService` + `react-router` and was out of budget. Deferred.

## Linked Issue

- https://linear.app/evoai/issue/EVO-1258


dpaes:
## 🔍 Code Review (EVO-1258)

Review of the autosave / dirty-state refactor. **Outcome: 2 blockers (HIGH) to fix before merge**, plus 2 MEDIUM and 4 LOW.

The **store** design is solid: `useFlowEditorStore` (Zustand) as the single source of truth, the `idle→dirty→saving→error` state machine, the 5s debounce coordinated with manual save, the atomic `commitSave`, the 30s retry gated by `retryScheduled`, and the IDB/localStorage helpers with graceful fallbacks + README. The 48 tests cover the store and helpers **in isolation**.

⚠️ Both blockers live in the **consumer↔canvas wiring** — exactly where the unit tests don't reach (no test mounts `BaseFlowCanvas`/`JourneyFlowEditor`). The PR body discloses this gap ("No full mount-level integration test"), so a green 48/48 doesn't refute either finding.

---

### 🔴 BLOCKER H1 — "Recover version" never reaches the canvas → **AC#1 broken**

`acceptRecovery()` updates only the store; the canvas never receives the recovered nodes.

- `BaseFlowCanvas.tsx:175-176` — `useNodesState(initialNodes)` seeds state **once** at mount; there is **no `useEffect` re-syncing `initialNodes → setNodes`**. The canvas is uncontrolled after mount.
- `useFlowEditorStore.ts:305-323` (`acceptRecovery`) — touches the store only; no `setJourney`, no remount signal.
- `JourneyFlowEditor.tsx:809-813` — the "Recover" button only calls `acceptRecovery()`.

**Repro:** edit → kill the tab → reopen → click "Recover" → **the canvas still shows the server version**. From there: if the user interacts, the first `select` change fires `setFlow` and overwrites `currentSnapshot` back to server-derived nodes (recovered data lost); if the 5s autosave fires first, it PUTs the recovered snapshot the user can't see on screen. The AC's "restores on accept" never happens visually.

**Fix direction:** accept must also `setJourney(recovered)` **and** force `BaseFlowCanvas` to re-seed (e.g. a `key` that changes on remount), or make the canvas controlled.

### 🔴 BLOCKER H2 — Selecting/measuring a node flips "dirty" → **Pain #9 (c) not resolved**

`onFlowDataChange` fires on **every** `NodeChange`, and the diff uses `JSON.stringify` over the full xyflow node array (which includes `selected`, `measured`, `position`).

- `BaseFlowCanvas.tsx:248-250` — `onFlowDataChange(updatedNodes, edges)` is called unconditionally for any change (select, dimensions, position).
- `useFlowEditorStore.ts:117-121` — `snapshotsEqual` compares the whole snapshot via `JSON.stringify`, with no volatile-field filtering.

**Repro:** open a saved journey → **click a node to open its config panel** → the header jumps to "Unsaved changes" and arms autosave, with no real edit.

**Self-pollution aggravator:** `saveChanges` PUTs `snapshot.nodes` verbatim (`JourneyFlowEditor.tsx:582`), so `selected:true` is **persisted**. On the next open the node comes back with `selected` baked in; the canvas mounts without selection → diff → **dirty on open, before any input**. This hits exactly what the story set out to fix ("clear line between local edit and persisted"). Side effect: a `toast.success` on every spurious autosave (L1).

**Fix direction:** filter non-substantive changes (`select`/`dimensions`) before marking dirty, or normalize the snapshot (strip `selected`/`measured`) before both `snapshotsEqual` and save.

---

### 🟡 M1 — `variables` tracked in dirty/atomic/recovery but excluded from the save payload
- `JourneyFlowEditor.tsx:526` — `serverSnapshot.variables` is **hardcoded `[]`** (never hydrated from the server).
- `JourneyFlowEditor.tsx:582` — the save builds `flowData = { nodes, edges }`, **no `variables`**.
- `useFlowEditorStore.ts:233-273` — `commitSave` sets `serverSnapshot = syncedSnapshot` (with variables) → the store believes variables it never sent are "synced".

Since variables are persisted independently via `useJourneyVariables` (`EnvironmentManager.tsx:131`), **no user data is lost** — but the single source of truth conflates two flows: the store tracks variables without ever hydrating or saving them through the same path → inconsistent state and a **false recovery prompt** when only variables differ.

### 🟡 M2 — In-app navigation guard only on the "Back" button (already disclosed)
`requestNavigate` (`JourneyFlowEditor.tsx:673`) is only used by `handleBack`. SPA navigation via menu/sidebar escapes the unsaved-changes dialog. `beforeunload` covers tab close/refresh (AC#5 ✅) but not in-app route changes.

---

### 🟢 LOW
- **L1** `toast.success` on every autosave (`JourneyFlowEditor.tsx:641`) — noisy; should mostly disappear once H2 is fixed.
- **L2** 30s retry with no backoff/cap (`useFlowEditorStore.ts:285-302`) — infinite loop while the server is down.
- **L3** `flowData` is a fresh object every render (`JourneyFlowEditor.tsx:737`) → fires the `BaseFlowEditor.tsx:124-129` effect on every re-render (harmless churn).
- **L4** Reverting an edit back to the server state leaves the autosave/IDB timers armed; benign because `saveChanges` bails on `status === 'idle'`.

---

**Validation note:** I did not run the suite locally — the new deps (`idb-keyval`, `fake-indexeddb`) aren't installed in the working tree. The blockers are structural and independent of that. i18n keys verified in pt-BR (10 present).

H1 and H2 are the blockers; the rest is at the maintainer's discretion.

dpaes (approved):
## ✅ Re-review — APPROVED

Round-2 review, verifying each round-1 finding against the actual code at the PR head (`ca2dba8`), not just the writeup.

### Round-1 blockers — both correctly resolved

**H1 — recovery never reached the canvas.** Fixed. `acceptRecovery` bumps `recoveryEpoch` in the store and `JourneyFlowEditor` applies `key={flow-canvas-${recoveryEpoch}}` to `BaseFlowEditor`, forcing a remount. Confirmed `BaseFlowCanvas` seeds `useNodesState(initialNodes)` **only on mount** (no re-sync effect), so the remount was genuinely required. The key piece in `ca2dba8`: the `flowData` memo now reads from `store.currentSnapshot` (not `journey.flowData`), so the remounted canvas picks up the recovered nodes. The chain closes → **AC#1 satisfied**.

**H2 — selecting/measuring a node marked dirty.** Fixed at the store layer rather than the canvas: `handleNodesChange` still fires `onFlowDataChange` on every `NodeChange`, but `snapshotsEqual` now normalises both sides via `stripVolatileNodeFields` (drops `selected`/`measured`/`dragging`), so a click is a no-op for dirty detection. `saveChanges` also PUTs through `normalizeNodesForPersist`, so the self-contamination loop (`selected:true` persisted → dirty on reopen) is gone. Regression test covers the contaminated-payload case. **Pain #9(c) resolved**.

### Medium / low
- **M1 (variables in the store)** — resolved. `FlowSnapshot` is now `{nodes, edges}` only; `setVariables` removed; `useJourneyVariables` remains the single owner. No orphan references.
- **L1 (toast on every autosave)** — resolved via the `silent` flag.
- **L2 (uncapped retry)** — resolved. Backoff 30s→60s→120s, `MAX_AUTO_RETRIES=3`, banner i18n parameterised with `{{seconds}}` across all 6 locales (verified). Auto-retry does not reset the budget, manual does — correct.
- **L3 (flowData identity churn)** — resolved via `useMemo` + memoised `initialEdges`.
- **Bonus (config-panel edits not marking dirty)** — `BaseFlowCanvas.updateNode` now notifies `onFlowDataChange`. Nice catch from the smoke test.
- **M2 (SPA nav guard)** and **L4 (timer/IDB armed on revert)** — deferred with reasonable justification.

### Non-blocking follow-ups (all LOW)
1. **L4 is deeper than the deferral assumes.** The "benign because `saveChanges` aborts on idle" note only covers the autosave-timer half. The other half: reverting an edit back to *exactly* the server state does not clear the IDB snapshot already written by the prior dirty edit → on reopen, `loadSnapshot` returns the stale dirty payload → `divergesFromServer=true` → **false recovery prompt**. No data loss. Optional one-liner: `clearSnapshot` in the revert-to-idle branch of `applyEdit`.
2. **`updateNode` calls `onFlowDataChange` inside the `setNodes` updater** (`BaseFlowCanvas.tsx:392-412`) — a side effect in a reducer; StrictMode double-invokes it in dev. Functionally safe (the store's `applyEdit` is idempotent), but a code smell.
3. **Stale comment** at `useFlowEditorStore.ts:212` still mentions `setVariables` (removed in M1). Trivial.
4. **Coverage gap persists:** no test mounts `JourneyFlowEditor`/`BaseFlowCanvas`. The 14 new tests cover the store primitives well; the H1 remount wiring and the `updateNode` bonus rely on code-read + manual smoke.

### Validation caveat
I did not run the suite locally — `idb-keyval`/`fake-indexeddb` aren't installed in the working tree and `pnpm` isn't on PATH. I verified the store logic and the spec by reading (consistent and correct); the reported 62/62 is the author's.

---

Both blockers are closed correctly and the deferrals are reasonable. Approving and merging into `develop`.

PR #114, opened 2026-05-28
feat(segments): drag-drop canvas builder + preview (EVO-1247)

daniloleonecarneiro:
## Summary
Extends the existing segments feature (no duplication):

- **Preview**: debounced (500ms) count + sample, with 422 keep-work (never mutates the form). Wired into `SegmentCreateEdit`.
- **Drag-drop canvas builder** (`@xyflow/react` via `BaseFlowEditor`): 8-node palette (Everyone, UserProperty, CustomAttribute, Label, Performed, RandomBucket, Email, Channel), reusing the existing `SegmentConditionEditor` as the node inspector, with an And/Or combinator. Toggled **Form ⇄ Canvas** over the same definition state.
- Routes segment CRUD + preview through the **CRM proxy** (`api → /api/v1/segments`; `update` via **PUT**).
- Adds the `ChannelNode` type to the union and en/pt-BR i18n keys.

## Test plan
- `vitest run` segments specs → 8 passing (palette = 8 types, adapter roundtrip/Everyone/empty/minimal, preview service via proxy).
- `tsc` exit 0; eslint clean on new files.

## Notas
- Parte da EVO-1247. Depende do proxy no **evo-ai-crm-community** (PR #103) e do `POST /segments/preview` no **evo-flow** (PR #14). Ordem de merge: evo-flow → crm → frontend.
- Reusa o RBAC `segments` e as rotas `/settings/segments` existentes.

PR #106, opened 2026-05-25
feat(frontend): redesign Flow Builder palette + wire assign-bot + 4 categories (EVO-1268)

nickoliveira23:
## Summary

EVO-1268 closes Pain #2 of discovery 8.1 (Journey Components palette). Five concrete changes:

- **assign-bot ghost wired** (EVO-1259 audit G1). Component + panel + pt-BR strings shipped by EVO-1260 were unreachable: barrel, JourneyFlowEditor imports, nodeTypes registry, miniMap, renderConfigPanel switch and palette entry all missing. Closed. Backend handler stays as EVO-1262 work — the node will silently no-op at runtime until that lands, same as 5 sibling nodes already in palette (G3).
- **Categories consolidated to 4**: Control Flow, Communication, Contact, Conversation. Renamed `actions -> controlFlow`, dropped the standalone `labels` bucket (add-label/remove-label now live under Contact). Triggers stays seeded (single journey-trigger-node); Pipeline category not introduced until EVO-1265/66/72/73 ship pipeline nodes.
- **Grouped layout in the 'All' view** (AC#1). When the user picks 'All Components' with no active search, the panel renders one `<section>` per category with an ALL-CAPS header + node count, instead of a flat list. Search collapses back to flat list across categories. Specific category selection also stays flat (single bucket).
- **searchKeywords on every node**. New optional `string[]` field on the `NodeType` interface; the palette filter substring-matches against keywords in addition to the locale-resolved name + description. Populated with 5-7 English aliases per node (e.g. `wait` finds via 'delay/pause/sleep/timer', `add-label` via 'tag/etiqueta/mark', `assign-agent` via 'user/operator/handoff').
- **Palette UX polish**. Category dropdown (`SelectTrigger`) bumped from `h-10` to `h-auto min-h-14` with `py-2.5 px-3` so the two-line label+description breathes. Scroll container now uses `max-h-[60vh]` (was `max-h-96`) and gets a `pt-4` so 8-10 nodes are visible on a typical viewport without scroll (AC#1 d). Section headers get top padding so the first one is not flush against the search panel border.

## Security

n/a — pure UI / theming / drag-and-drop wiring. No new routes, no data paths touched.

## Test plan

- `evo-ai-frontend-community: pnpm exec tsc -p tsconfig.app.json --noEmit` — clean (only the pre-existing `EvolutionHubConfig` error survives, also present on develop).
[...]

PR #116, opened 2026-05-29
fix(contacts): mostrar traits na timeline de eventos + CSP localhost (EVO-1571)

daniloleonecarneiro:
## Summary
Dois fixes da feature de Histórico/Tracking (aba "Eventos" do contato):

1. **`fix(contacts)`** — `ContactEventCard` agora renderiza `traits` além de `properties`. Eventos `identify` (`contact.updated`, `contact.custom_attribute.changed`) têm `properties: {}` e os dados reais (nome, `customAttributes`, `changes`) em `traits`; antes o card mostrava `{}` e o evento ficava sem nenhuma informação útil. Fallback "no additional data" só quando ambos vazios.
2. **`fix(docker)`** — corrige o `connect-src` do CSP no `docker-entrypoint.sh`: o `sed` que libera `http://localhost:*` em dev não casava com o CSP real do `nginx.conf` (faltava `blob:`), bloqueando login (`:3001`) e chamadas ao CRM/evo-flow no browser.

## Test plan
- `npm run build` (tsc + vite)
- E2E: editar um contato → aba Eventos → expandir o `contact.updated` → ver os dados (traits: nome, changes, custom attributes).
- Login e chamadas a `http://localhost:*` não são mais bloqueados pelo CSP.

Relacionado: EVO-1571

## Summary by Sourcery

Fix contact event timeline data visibility and align CSP configuration for localhost in development.

Bug Fixes:
- Ensure contact event cards display traits as well as properties so identify-style events show their payload instead of appearing empty.
- Show a clear fallback message when an event has neither properties nor traits data to render.
- Correct the development Content Security Policy connect-src directive so localhost HTTP calls are allowed alongside existing sources.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -27,7 +27,7 @@ APP_ENV="${VITE_APP_ENV:-development}"
 
 if [ "$APP_ENV" = "development" ]; then
   # Development: Allow localhost connections and permissive frame-ancestors for widget
-  sed -i "s|connect-src 'self' https: wss: ws:|connect-src 'self' https: wss: ws: http://localhost:*|g" /etc/nginx/conf.d/default.conf
+  sed -i "s|connect-src 'self' blob: https: wss: ws:|connect-src 'self' blob: https: wss: ws: http://localhost:*|g" /etc/nginx/conf.d/default.conf
   sed -i "s|frame-ancestors 'self'|frame-ancestors *|g" /etc/nginx/conf.d/default.conf
 fi
 
--- a/src/components/contacts/ContactEventCard.tsx
+++ b/src/components/contacts/ContactEventCard.tsx
@@ -55,6 +55,14 @@ function ContactEventCardImpl({ event }: ContactEventCardProps) {
 
   const enriched = event.enriched;
 
+  // Identify events (e.g. contact.updated) carry their payload in `traits`,
+  // not `properties` (which is `{}`). Track/page events use `properties`.
+  // Render whichever has data so no event shows up empty.
+  const hasKeys = (o?: Record<string, unknown>): boolean =>
+    !!o && typeof o === 'object' && Object.keys(o).length > 0;
+  const hasProps = hasKeys(event.properties);
+  const hasTraits = hasKeys(event.traits);
+
   return (
     <Card className="border border-border bg-card transition-colors hover:bg-accent/30">
       <CardContent className="p-4">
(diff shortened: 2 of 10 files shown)


dpaes (changes requested):
## Review (EVO-1571 · 7.8 traits + CSP) — change requested

Thanks for this — the CSP fix and the traits rendering are both correct. Two things before merge.

### 🔴 Blocker — an existing test now fails

`src/components/contacts/ContactEventCard.tsx` moves `id={propertiesId}` from the `<pre>` to the new wrapper `<div>`. The existing spec asserts the id lives on the `<pre>`:

`src/components/contacts/__tests__/ContactEventCard.spec.tsx` → **"exposes aria-controls linking the button to the properties pre element"**:

```ts
const pre = await screen.findByText((_, node) => node?.tagName.toLowerCase() === 'pre');
expect(pre.id).toBe(controls); // now fails: <pre> no longer has the id
```

Since `aria-controls` on the button still points at `propertiesId` (now the `<div>`), `pre.id` is `""` and the assertion fails. Please either keep the id on the controlled element the test expects, or update the spec to assert the controlled region is the wrapper `<div>` (the more correct a11y semantics now that there can be two `<pre>` blocks).

⚠️ Heads-up: this repo's CI runs Sourcery only — vitest is **not** executed in CI, so this lands red without anything flagging it. Please run `npx vitest run src/components/contacts/__tests__/ContactEventCard.spec.tsx` locally.

### 🟡 Non-blocking — i18n: new labels are English in every non-EN locale

`events.card.traits` and `events.card.noData` only exist as `defaultValue` in the component; they are not in any locale file. So pt-BR/pt/es/fr/it users see "Traits" and "No additional data for this event" in English. (`events.card.properties` is already translated — that one is fine.)

Please add both keys to all six locales (`en`, `pt-BR`, `pt`, `es`, `fr`, `it` `contacts.json`) and to the `evo1244Keys` list in `src/i18n/locales/contacts-parity.spec.ts` so the parity test covers them.

### ✅ Confirmed correct
- CSP: the new `sed` pattern `connect-src 'self' blob: https: wss: ws:` now matches `nginx.conf` line 25 exactly (the old pattern was missing `blob:`, which is why it never matched). 
- Traits rendering: renders `properties` and/or `traits`, with the "no data" fallback when both are empty; `traits` is present on the `ContactEvent` type.

dpaes:
## ✅ Re-review (round 2) — both round-1 findings resolved

- **aria-controls test (was the blocker):** fixed. The `id` moved to the wrapper `<div>` and the spec now asserts `region.contains(pre)` + `within(region).getByText(/foo/)`. Verified it passes — `makeEvent` sets no `traits` by default, so the test case renders a single `<pre>` and `findByText(pre)` stays unambiguous.
- **i18n leak (was non-blocking):** `events.card.traits` / `events.card.noData` added to all 6 locales (en/es/fr/it/pt-BR/pt) and to the `evo1244Keys` parity list. Confirmed the parity spec imports exactly those 6 locales, so parity holds.
- **Type-safe:** `event.traits` is valid — `src/types/contacts/contactEvent.ts` already declares `traits?`. The CSP fix is unchanged and correct (matches `nginx.conf`).

⚠️ FE CI doesn't run vitest/tsc — I verified by static reading; please run `vitest run` + `tsc` locally before merge.

No remaining issues on this PR. I'm **holding the merge** only because the EVO-1571 card is bouncing back as a unit pending #114 (broken submodule pointers). This PR is independently mergeable (pure FE render; #106 is already on `develop`) — say the word and I'll merge it on its own.

dpaes (approved):
## Re-review round 3 (EVO-1571 · 7.8 traits + CSP) — approved ✅

Re-reviewed against the current head (`d0e6d52`) and the real repo state. The two round-1 findings are resolved; no pending issues.

### ✅ aria-controls test (was the 🔴 blocker)
The `id={propertiesId}` moved from the `<pre>` to the wrapper `<div>`, and the spec now asserts `region.contains(pre)` + `within(region).getByText(/foo/)`. Verified the `makeEvent` helper does **not** set `traits` by default (only `properties: {}`), so the test renders a **single** `<pre>` — the `findByText(node => node.tagName === 'pre')` lookup stays unambiguous. The other `<pre>` test (redaction) is properties-only and asserts on content, not on the id/structure, so it is unaffected by moving the id to the wrapper.

### ✅ i18n leak (was 🟡)
`events.card.traits` and `events.card.noData` are present in all **6** locales (en/es/fr/it/pt-BR/pt) and added to the `evo1244Keys` list in `contacts-parity.spec.ts`. Confirmed there are exactly 6 locale dirs, so the parity spec covers them.

### ✅ Type safety & CSP
- `src/types/contacts/contactEvent.ts` declares `traits?: Record<string, unknown>`, so `event.traits` is type-safe (tsc-clean).
- `nginx.conf` (line 25) contains the exact string `connect-src 'self' blob: https: wss: ws:`, so the updated `sed` pattern matches and appends `http://localhost:*` in dev. The old pattern (missing `blob:`) was a silent no-op — this is the correct fix.

### ⚠️ Pre-merge note (non-blocking)
FE CI runs Sourcery only (no vitest/tsc). Since the round-1 blocker was a failing test and this PR edits that exact spec, please run `vitest run` + `tsc` locally before merge as a final check. Logic verified by reading; approving.

PR #120, opened 2026-05-29
fix(frontend): propagate canvas edge mutations to the editor store (EVO-1573)

nickoliveira23:
## Summary

`BaseFlowCanvas.handleConnect` and `handleEdgesChange` updated xyflow's internal state via `setEdges` / `onEdgesChangeInternal` but never called the parent-supplied `onFlowDataChange` / `onFlowDataChangeExtended` callbacks. The Customer Journey editor store (`useFlowEditorStore`) consequently saw a fresh connection only when an incidental node-change carried the updated `edges` closure with it. A user who connected two nodes and clicked **Save** without moving anything afterwards silently lost the edge on reload (Pain #9 (a) on the discovery 8.1, surfaced during EVO-1265 smoke).

This PR closes that gap and adds review-driven hardening.

## Changes

- **`handleConnect`**: compute `updatedEdges = addEdge(edge, edges)` from the closure, call `setEdges(updatedEdges)` with a **bare value (NOT a functional updater)**, then fire `onFlowDataChange` and `onFlowDataChangeExtended`. The pure setter avoids React 18 StrictMode double-firing the callbacks, and matches the shape of `handleEdgesChange` / `handleNodesChange` already in the file.
- **`handleEdgesChange`**: filter `select`-type changes before propagating. Edge selection is volatile UI state and the store has no `stripVolatileEdgeFields` (mirror of `stripVolatileNodeFields` from `useFlowEditorStore.ts:172`), so without the filter a selection click would mark the journey dirty. xyflow still receives all changes via `onEdgesChangeInternal` so selection visuals are preserved; only `add` / `remove` / `replace` reach the parent via `applyEdgeChanges(persistChanges, edges)`.
- **`BaseFlowCanvas.spec.tsx`** (new): 5 tests cover `handleConnect` propagation (AC1), `handleConnect` via `onFlowDataChangeExtended`, `handleEdgesChange` delete (parity), connect-then-node-move workaround (AC3 — no-regression of the path users had been relying on), and select-only-change does-NOT-dirty (review-finding M1). Vitest mocks `@xyflow/react`'s `ReactFlow` component to capture handler props while leaving the real `useNodesState`, `useEdgesState`, and `addEdge` hooks live.

## AC mapping (from the Linear card)

- **AC1** — connect → store dirty → save payload contains edge: covered by spec tests 1 & 2.
- **AC2** — save + reload persists the edge: implied by AC1 contract (frontend correctness) + manual smoke during EVO-1265.
- **AC3** — connect-then-move workaround path unchanged: covered by spec test 4.
- **AC4** — spec under `src/components/base/BaseFlowCanvas` covers the connect-without-move path: this PR.

## Security

- No new attack surface. Only changes are to in-memory React state propagation and a vitest mock.

## Test plan

- `pnpm exec tsc -b --noEmit` — exit 0
- `pnpm exec vitest run src/components/base/BaseFlowCanvas` — 5/5
- `pnpm exec vitest run src/components/journey src/components/base src/services/automation` — 122/122 (1 pre-existing fail on `tokens.contrast.spec.ts` is missing `colorjs.io` on develop, unrelated)
- Manual smoke recommended after merge: open `/journey/:id/flow`, drag two nodes, connect them, click Save without moving anything, reload → edge persists.

## Changed Files

- `src/components/base/BaseFlowCanvas.tsx`
- `src/components/base/BaseFlowCanvas.spec.tsx` (new)

## Linked Issue

- EVO-1573 — https://linear.app/evoai/issue/EVO-1573

### DIFF
--- a/src/components/base/BaseFlowCanvas.spec.tsx
+++ b/src/components/base/BaseFlowCanvas.spec.tsx
@@ -1,0 +1,227 @@
+import { act, render } from '@testing-library/react';
+import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
+
+beforeAll(() => {
+  if (!window.matchMedia) {
+    window.matchMedia = (query: string) =>
+      ({
+        matches: false,
+        media: query,
+        onchange: null,
+        addListener: () => {},
+        removeListener: () => {},
+        addEventListener: () => {},
+        removeEventListener: () => {},
+        dispatchEvent: () => false,
+      }) as unknown as MediaQueryList;
+  }
+});
+
+const reactFlowMocks = vi.hoisted(() => ({
+  capturedProps: { current: null as Record<string, unknown> | null },
+}));
+
+vi.mock('@xyflow/react', async importOriginal => {
+  const actual = await importOriginal<typeof import('@xyflow/react')>();
+  return {
+    ...actual,
+    ReactFlow: vi.fn((props: Record<string, unknown>) => {
+      reactFlowMocks.capturedProps.current = props;
+      return (
+        <div data-testid="react-flow-mock">
+          {(props.children as React.ReactNode) ?? null}
+        </div>
+      );
+    }),
+  };
+});
+
+import { ReactFlowProvider } from '@xyflow/react';
+import { BaseFlowCanvas } from './BaseFlowCanvas';
+import { DnDProvider } from '@/contexts/DnDContext';
+import { DarkModeProvider } from '@/contexts/ThemeContext';
+
+const NoopNode = () => <div />;
+
+function renderCanvas(opts: {
+  onFlowDataChange?: ReturnType<typeof vi.fn>;
+  onFlowDataChangeExtended?: ReturnType<typeof vi.fn>;
+  initialNodes?: Array<Record<string, unknown>>;
+  initialEdges?: Array<Record<string, unknown>>;
+}) {
+  const initialNodes = opts.initialNodes ?? [
+    { id: 'n1', type: 'noop', position: { x: 0, y: 0 }, data: {} },
+    { id: 'n2', type: 'noop', position: { x: 200, y: 0 }, data: {} },
+  ];
+  const initialEdges = opts.initialEdges ?? [];
+
+  return render(
+    <DarkModeProvider>
+      <DnDProvider>
+        <ReactFlowProvider>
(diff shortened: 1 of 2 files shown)


dpaes (approved):
**Approved** ✅ — EVO-1573 (canvas edge propagation).

Correct root-cause fix: `handleConnect` / `handleEdgesChange` now propagate edge mutations to `onFlowDataChange` / `onFlowDataChangeExtended`, mirroring the existing `handleNodesChange` path, so a freshly drawn (or deleted) edge reaches the editor store and the next save snapshot includes it — closing the silent data-loss on reload.

**Verified during review**
- **No double-fire:** traced xyflow `12.10.0` — edges are controlled (`edges=`, not `defaultEdges`), so a connection invokes only `onConnect`, never `onEdgesChange('add')`. One connect → one `onFlowDataChange`.
- **Blast radius contained:** only Journey (intended fix, no regression) and Segments consume the canvas. Segments is a definitional no-op (`SegmentCanvasNode` has no handles, `flowFromDefinition` seeds `edges:[]`), so the new paths never run there. No Campaigns/Automation canvas consumer.
- **Spec is genuine:** the mock overrides only `ReactFlow`, leaving real `addEdge`/`applyEdgeChanges`/`useEdgesState`; tests 1-3 genuinely fail pre-fix.
- The `select`-filter (M1) and the `handleEdgesChange` delete parity are good additions.

**Non-blocking notes (dev discretion)**
- **AC2 (save + reload persistence)** has no automated coverage (not unit-testable) and no fix-verification smoke — the referenced "EVO-1265 smoke" is what *surfaced* the bug, not a re-test. A quick reload smoke is worth it since the whole symptom is reload data-loss.
- The spec header `// (each test fails on the pre-fix develop)` is inaccurate for tests 4 & 5 — both pass pre-fix; they guard no-regression and the new select-filter, not the pre/post delta.
- Edge volatile fields aren't normalized at persist (a `selected:true` edge can ride into the payload) — pre-existing; the select-filter is a partial mitigation. A `stripVolatileEdgeFields` mirror would fully meet the stated parity goal.
- Comment says "React 18" but the project runs React `19.2.1` — substance correct, label only.

_Note: frontend CI runs Sourcery only — tsc/vitest (122/122) are author-reported (verified by inspection). Merging squash; umbrella submodule pointer intentionally not bumped (batched at release)._

PR #139, opened 2026-06-05
fix(frontend): apply account language change instantly without reload (EVO-1636)

nickoliveira23:
## Summary

`/settings/account` persisted the chosen language to the backend but never switched the running SPA, so the new language only took effect after a manual F5 (EVO-1636).

- `handleGeneralSubmit` now calls the existing `useLanguage().changeLanguage` after a successful `updateAccount` — it runs `i18n.changeLanguage` (reactive re-render via react-i18next, `useSuspense:false` + default `bindI18n`) and persists to `localStorage`.
- The normalized locale is computed once and reused for both the API payload and the language switch, so the persisted value and the applied value cannot drift.
- Placed inside the `try`, after the `await`, so the UI does not switch if the save fails.

## Security
N/A — client-side UI behavior only.

## Test plan
- `pnpm exec eslint src/pages/Customer/Settings/Account/AccountSettings.tsx` — no new issues
- Manual smoke: `/settings/account` → change language → Save → UI switches instantly, no F5; selection persists after a real reload.

## Changed Files
- `src/pages/Customer/Settings/Account/AccountSettings.tsx`

## Linked Issue
- EVO-1636

## Summary by Sourcery

Apply language changes in the account settings page immediately after saving, keeping the persisted and active locales in sync.

New Features:
- Switch the SPA language instantly on successful account settings save based on the selected locale.

Bug Fixes:
- Prevent mismatches between the locale saved to the backend and the locale applied in the UI by reusing a normalized locale value.

### DIFF
--- a/src/pages/Customer/Settings/Account/AccountSettings.tsx
+++ b/src/pages/Customer/Settings/Account/AccountSettings.tsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { useLanguage } from '@/hooks/useLanguage';
+import { type Locale } from '@/i18n/config';
 import {
   Input,
   Label,
@@ -52,7 +53,7 @@ function SectionLayout({
 }
 
 export default function AccountSettings() {
-  const { t } = useLanguage('accountSettings');
+  const { t, changeLanguage } = useLanguage('accountSettings');
   const { can, isReady: permissionsReady } = useUserPermissions();
   const normalizeAccountLocale = (locale?: string | null): string => {
     if (!locale) return 'pt-BR';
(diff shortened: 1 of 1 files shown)


dpaes (approved):
## Code review — EVO-1636 (approved ✅)

Reviewed in context (branch checked out), verifying the contracts, not just the diff.

### AC coverage
- **AC1 (UI updates with no F5):** `changeLanguage(normalizedLocale)` calls `i18n.changeLanguage(lng)` → react-i18next re-renders reactively (`useSuspense:false`, default `bindI18n`; all 6 locales bundled eagerly so there's no async resource load). Smoke-tested by the author.
- **AC2 (persists after reload):** the same `changeLanguage` does `localStorage.setItem('i18nextLng', lng)`, which the initial-language detection reads on boot.
- **AC3 (reload fallback):** not needed — the reactive path works.
- **Placement is correct:** inside `try`, **after** the `await updateAccount` resolves and **before** the toast / `loadAccountData`. If the save throws, the language does not change → consistent state. Nice side effect: the success toast already renders in the new language.
- `Locale` is exported (`i18n/config.ts:335`) and imported as a `type`; `tsc -b` reported clean.

### Single non-blocking note (already disclosed by the author, L1)
`normalizeAccountLocale` returns `string`, not `Locale`, so `as Locale` is an unchecked cast (it only maps `pt_BR`→`pt-BR`; any other value passes through). In practice the language `<select>` is bounded to the 6 supported locales, and an unsupported value would fall back to English via `fallbackLng` — only `localStorage` would keep the bad value. Pre-existing pattern (the initial load at line 133 already uses the same helper); not a regression. Optional hardening: have `normalizeAccountLocale` return a validated `Locale`.

Minimal, surgical fix, all 3 ACs met, no side effects. Approving and squash-merging.

PR #141, opened 2026-06-05
fix(automation): form UUID schemas, submit gate, required markers + i18n (EVO-1639)

daniloleonecarneiro:
## EVO-1639 — Form de automação: schemas UUID, botão travado, UX/i18n

**Fix:**
- Schemas zod aceitam IDs **UUID** (`assign_team`/`assign_agent`/`team_ids`/`inbox_id`/`assigned_to_id`) — antes `number` travava o submit; inputs preservam UUID.
- Botão 'Criar' via `safeParse(watch())` (o `formState.isValid` não recomputava com `useFieldArray`).
- `*` em obrigatórios, '(opcional)' em condições; i18n `automation.json` p/ es/fr/it/pt.
- Hardening: `getEventLabel` não crasha com locale nullish (corrige baseline do `AutomationForm.spec` quebrado no develop — EVO-1263).

**Pendente (follow-up):** `send_email_to_team` precisa de seletor de time na UI.

**Testes:** suite de automação **104 verdes**; tsc limpo.

Parte da EVO-1637.

### DIFF
--- a/src/components/automation/ActionRow.tsx
+++ b/src/components/automation/ActionRow.tsx
@@ -365,16 +365,15 @@ function ActionParamsRenderer({ control, index, actionName, formData, t }: Param
                 </div>
                 <div className="grid grid-cols-2 gap-2">
                   <Input
-                    type="number"
                     value={
                       current.assigned_to_id != null && current.assigned_to_id !== ''
                         ? String(current.assigned_to_id)
                         : ''
                     }
                     onChange={(e) => {
-                      const raw = e.target.value;
-                      const num = raw === '' ? undefined : Number(raw);
-                      setField('assigned_to_id', Number.isNaN(num) ? undefined : num);
+                      // user ids are UUID strings — keep the raw value, don't coerce to Number
+                      const raw = e.target.value.trim();
+                      setField('assigned_to_id', raw === '' ? undefined : raw);
                     }}
                     placeholder={t('form.fields.actionRow.params.create_pipeline_task_assignee')}
                   />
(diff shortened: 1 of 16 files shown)


dpaes (approved):
## Review — EVO-1639 · ✅ Approved

Reviewed against the EVO-1637 acceptance criteria. The form fixes are correct, well-scoped, and carry regression tests for the UUID bug class.

**What's solid**
- `entityIdSchema = z.union([z.string().min(1), z.number()])` for `team_ids` / `assign_team` / `assign_agent` / `inbox_id` / `assigned_to_id` correctly fixes the UUID rejection. It's a strict superset of the old `number` schema, so legacy numeric ids still validate — edit mode is safe. Regression specs cover it.
- `canSubmit = automationRuleSchema.safeParse(watch()).success` is the right gate: RHF's `formState.isValid` is unreliable with nested `useFieldArray`. It's a pure synchronous derivation (no setState → no render loop) and mirrors the exact onSubmit check.
- `getEventLabel` nullish-locale hardening is strictly safer than the previous `.toString()`.
- i18n es/fr/it/pt are genuinely localized (not EN copies), 185 keys at full parity with `en`, registered in `config.ts`; no namespace collision with the pre-existing pt-BR/en.

**Non-blocking (recommend fast-follow, not gating this merge)**
- ⚠️ `send_email_to_team` is a UX dead-end: the action is offered, its schema requires `team_ids.min(1)`, but there is no team-selector in the UI — so `team_ids` stays `[]`, the action is permanently schema-invalid, and because `canSubmit` aggregates all actions, adding it disables **Save** for the whole rule until it's removed. Recommend hiding the action until the selector ships (or shipping the selector). [pre-existing, already disclosed]
- The card mentions a CSP fix in `docker-entrypoint.sh` (`connect-src http://localhost:*`) that isn't in this diff — did it get left out / go to another branch?
- Pre-existing (not introduced here): the event dropdown exposes 7 events while the i18n lists 9 (`+conversation_resolved`, `+conversation_status_changed`); a rule on the 2 extra events would fail the resolver, but it already did pre-PR.

Caveat: frontend CI is Sourcery-only (no vitest gate) and the new submit-gate test is create-mode only, so "104 green / tsc clean" is self-reported, not CI-enforced.

Approving and merging to `develop`.

PR #143, opened 2026-06-05
fix(frontend): persist all Flow Builder editor mutations to the store (EVO-1643)

nickoliveira23:
## Summary

Flow Builder mutations that went through direct xyflow setters (`setNodes`/`setEdges`/`addNodes`) never notified the editor store (`useFlowEditorStore`), so the save snapshot omitted them — changes were lost on reload. Same root cause as EVO-1573 (edges), now closed for every editor mutation the Journey uses.

Fixed paths (each verified manually + by regression test):
- **Drop** an action node from the palette — `BaseFlowCanvas.handleDrop` default branch now fires `onFlowDataChange`/`onFlowDataChangeExtended`.
- **Delete node** (context menu) — centralized in `handleDeleteNode` (removes the node + its connected edges) with store notification.
- **Duplicate node** (context menu) — centralized in `handleDuplicateNode`.
- **Delete edge** (trash button) — `handleDeleteEdge` now propagates; `BaseDefaultEdge` uses the canvas-provided handler (passed via edge `data`) instead of its own `setEdges`.

`BaseFlowContextMenu` no longer mutates via `useReactFlow` — it delegates delete/duplicate to the parent through props. All handlers follow the EVO-1573 pattern (compute bare value → set → fire callbacks after, never inside the updater).

This unblocks the EVO-1634 executor wiring end-to-end: user-built journeys now actually persist their action nodes.

## Security
N/A — client-side editor state only.

## Test plan
- `pnpm test src/components/base/BaseFlowCanvas.spec.tsx` — 10/10 (5 EVO-1573 + 2 drop + 3 new: node delete / node duplicate / edge delete)
- `pnpm exec eslint <changed files>` — no new issues (net -2 pre-existing `any` removed)
- Manual smoke: drop / delete node / duplicate node / delete edge → Save → reload → all persist.

## Changed Files
- `src/components/base/BaseFlowCanvas.tsx`
- `src/components/base/BaseFlowContextMenu.tsx`
- `src/components/base/BaseDefaultEdge.tsx`
- `src/components/base/BaseFlowCanvas.spec.tsx`

## Linked Issue
- EVO-1643

### DIFF
--- a/src/components/base/BaseDefaultEdge.tsx
+++ b/src/components/base/BaseDefaultEdge.tsx
@@ -17,6 +17,7 @@ export default function BaseDefaultEdge({
   targetPosition,
   style = {},
   selected,
+  data,
 }: EdgeProps) {
   const { setEdges } = useReactFlow();
 
@@ -30,8 +31,17 @@ export default function BaseDefaultEdge({
     borderRadius: 15,
   });
 
+  // EVO-1643: prefer the canvas-provided handler (passed via edge `data`),
+  // which removes the edge AND notifies the editor store. The bare setEdges
+  // fallback keeps standalone usage working but does not persist on save.
   const onEdgeClick = () => {
-    setEdges((edges) => edges.filter((edge) => edge.id !== id));
+    const deleteEdge = (data as { handleDeleteEdge?: (edgeId: string) => void } | undefined)
+      ?.handleDeleteEdge;
+    if (deleteEdge) {
+      deleteEdge(id);
+    } else {
+      setEdges((edges) => edges.filter((edge) => edge.id !== id));
+    }
   };
 
   return (
--- a/src/components/base/BaseFlowCanvas.spec.tsx
+++ b/src/components/base/BaseFlowCanvas.spec.tsx
@@ -1,5 +1,6 @@
 import { act, render } from '@testing-library/react';
 import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
+import { useEffect } from 'react';
 
 beforeAll(() => {
   if (!window.matchMedia) {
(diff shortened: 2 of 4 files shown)


dpaes (approved):
## Review — EVO-1643 · ✅ Approved

Verified against the AC and the EVO-1573 mutation-propagation pattern. Clean fix for the whole class of "editor mutation bypasses the store → dropped on save".

**Verified**
- Every mutation now computes the bare value, calls `setNodes`/`setEdges` with it, then fires `onFlowDataChange`/`onFlowDataChangeExtended` **after** the setter (not inside the updater) — so StrictMode's double-invoke can't double-notify, matching the EVO-1573 fix. Drop, node-delete (+ connected edges), node-duplicate, and edge-delete are all covered.
- Edge-delete is wired end-to-end: `BaseFlowCanvas` injects `handleDeleteEdge` into `defaultEdgeOptions.data` (~line 649), and `BaseDefaultEdge`'s trash button prefers it, falling back to the bare `setEdges` only for standalone usage — so in the Journey/Flow Builder the trash button persists on save.
- `BaseFlowContextMenu` no longer mutates via `useReactFlow`; it delegates to the parent (`onDeleteNode`/`onDuplicateNode`), which is what routes the change through the store. The custom-`ContextMenuComponent` override path is unchanged (no regression; the default menu is the one the Flow Builder uses and it gets both handlers).
- `useCallback` deps include `nodes`/`edges`/`flowVariables`, so the handlers read fresh canvas state (no stale closure).
- +5 specs that fail on pre-fix `develop`.

Unblocks EVO-1634 end-to-end (a user-built Journey now persists all action nodes).

Caveat: frontend CI is Sourcery-only (no vitest gate), so "10/10 / tsc clean" is self-reported.

Approving and merging to `develop`.

PR #146, opened 2026-06-08
fix(frontend): paginate conversation list with current query and preserve total (EVO-1671)

nickoliveira23:
## Summary

Fixes the conversation list (`/conversations`) pagination: the header showed the correct API total (e.g. "81") on first load, but scrolling reset it to **0** and the infinite scroll stopped mid-list, so only a subset of conversations was ever shown.

Root cause was two compounding defects:
1. `loadMoreConversations` re-fetched with only `{ page }`, dropping the `status`/filters that produced the current list — page 2 hit `GET /conversations` unfiltered, a different query (and response shape) than page 1.
2. `APPEND_CONVERSATIONS` overwrote `conversationsPagination` wholesale, so a page whose meta normalized to `total: 0` zeroed the count and turned `has_next_page` false (the header count `?? visibleConversations.length` did not catch `0`).

## Fix
- Track the query that produced the list (`ConversationsQuery` = list params **or** a filter request) and **replay the same request** for the next page in load-more — `GET /conversations` for lists, `POST /conversations/filter` for advanced filters.
- `APPEND_CONVERSATIONS` now **preserves `total`/`total_pages`** when a later page reports `0`/missing and recomputes `has_next_page = page < total_pages`.
- Header count falls back to the visible length when `total` is `0`.

## Security
- Frontend-only; no endpoint or auth changes. Load-more now sends the same scoped query (status/filters) it already used for page 1.

## Test plan
- `frontend: pnpm exec tsc -b --noEmit`
- `frontend: pnpm exec eslint <changed files>` (0 errors)
- `frontend: pnpm exec vitest run src/contexts/chat/ConversationsContext.spec.tsx` (21 tests; +3 for APPEND pagination)
- Manual smoke: default view scrolls through all conversations with a stable count; active filter / quick-tab keeps the same filter across pages; search paginates within the search.

## Changed Files
- `src/types/chat/api.ts` — `ConversationsQuery` type
- `src/contexts/chat/ConversationsContext.tsx` — track current query; load-more replays it
- `src/contexts/chat/ConversationsContextReducer.ts` — APPEND preserves total/total_pages
- `src/contexts/chat/FiltersContext.tsx` — emit query descriptor on success
- `src/hooks/chat/useFilterHandlers.ts` — forward query descriptor to setConversations
- `src/types/chat/conversations.ts` — setConversations signature
- `src/components/chat/chat-sidebar/ChatSidebar.tsx` — count handles total 0
- `src/contexts/chat/ConversationsContext.spec.tsx` — APPEND pagination tests

## Scope note
This is the correctness fix. Scroll-transition smoothness (prefetch buffer + skeleton cushion on fast scroll / slow connections) is tracked separately as a follow-up to EVO-1617.

## Linked Issue
- EVO-1671

### DIFF
--- a/src/components/chat/chat-sidebar/ChatSidebar.tsx
+++ b/src/components/chat/chat-sidebar/ChatSidebar.tsx
@@ -995,8 +995,8 @@ const ChatSidebar = ({
         {/* Filter Actions */}
         <div className="flex items-center justify-between">
           <span className="text-sm text-muted-foreground">
-            {(conversations.state.conversationsPagination?.total ?? visibleConversations.length)}{' '}
-            {(conversations.state.conversationsPagination?.total ?? visibleConversations.length) === 1
+            {(conversations.state.conversationsPagination?.total || visibleConversations.length)}{' '}
+            {(conversations.state.conversationsPagination?.total || visibleConversations.length) === 1
               ? t('chatSidebar.conversation')
               : t('chatSidebar.conversations')}
           </span>
--- a/src/contexts/chat/ConversationsContext.spec.tsx
+++ b/src/contexts/chat/ConversationsContext.spec.tsx
@@ -396,4 +396,65 @@ describe('ConversationsContext reducer', () => {
       expect(matchesConversationId(b, '42')).toBe(true);
     });
   });
+
+  describe('APPEND_CONVERSATIONS pagination (EVO-1671)', () => {
+    const baseState = (total: number, page = 1, totalPages = 5) => ({
+      ...initialState,
+      conversations: [makeConversation({ id: 'c1', uuid: 'c1' })],
+      conversationsPagination: {
+        page,
+        page_size: 20,
+        total,
+        total_pages: totalPages,
+        has_next_page: page < totalPages,
+      },
+    });
+
+    it('preserves the previous total when the appended page reports total 0', () => {
+      const state = baseState(81, 1, 5);
+      const next = conversationsReducer(state, {
+        type: 'APPEND_CONVERSATIONS',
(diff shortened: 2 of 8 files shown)


dpaes (approved):
## Review — EVO-1671 · Approved

Solid fix — addresses all 3 documented root causes and the 4 ACs, verified against the code at the PR head (this repo's CI doesn't run vitest, so I checked the source, not the self-report).

- **loadMore replays the same query** via a `currentQueryRef` (a ref → no stale-closure): it records the page-1 query (`list` params or `filter` request) and replays it for the next page — GET with params, or re-`POST /conversations/filter` with `page+1`. All 3 `setConversations` callers forward the query. ✅ (AC1/AC2)
- **APPEND reducer preserves `total`/`total_pages`** when a later page reports 0/missing, and recomputes `has_next_page = page < total_pages` (the `page` fallback `incoming.page || (prev?.page ?? 1) + 1` parses correctly). ✅ (AC1/AC3/AC4)
- **Header `?? → ||`** so `total === 0` falls back to the visible length. ✅ (AC3)
- The 3 new reducer specs cover total-preservation + `has_next_page`.

Non-blocking (not gating merge):
- The advanced-filter load-more assumes `POST /conversations/filter` honours `page`. If the backend ignores it, load-more re-fetches page 1 (deduped, never advances). Backend is out of scope per the card — worth a one-line confirm that the filter endpoint paginates.
- The biggest behavioural change (query/filter replay) has no unit test — only the reducer does. Since CI doesn't run vitest, a loadMore-replays-the-filter test would be worth adding.

Approving and merging to develop.

PR #150, opened 2026-06-10
fix(frontend): strip doubled /api/v1 prefix from rbacService paths (EVO-1688)

nickoliveira23:
## Summary

`apiAuth`'s baseURL already ends in `/api/v1` (`src/services/core/apiAuth.ts:10`), but all **11** `rbacService` call sites passed `/api/v1`-prefixed paths — every RBAC request went out as `/api/v1/api/v1/...` and 404'd on the auth service.

**Live casualty:** the role dropdown in the add/edit user modal (`/settings/users` → `UserFormModal` → `useRoles()` → `fetchRoles`) came up empty, blocking user creation via the UI (the EVO-1040-style symptom). This PR repairs that path.

Fix: removed the redundant `/api/v1` from the 11 paths (the client keeps the prefix in its baseURL, matching the working siblings — `fetchRolesFull`'s `'/roles/full'` and the newer `rolesService`).

## Known limitation (verified against auth `routes.rb`, documented — out of scope)

5 of the 11 functions (`role_permissions` CRUD ×4, `resource_actions/for_permissions`) target routes the auth service **never exposed** — permissions are managed via `roles#bulk_update_permissions`, which the Admin Roles pages already consume through the newer `rolesService`. The only code referencing these 5 is dead (`hooks/rbac/usePermissions` has zero importers; 3 functions have no importers at all), so no live behavior depends on them. They stay as cleanup candidates.

## Validation

- `evo-ai-frontend-community: pnpm exec vitest run src/services/rbac/rbacService.spec.ts` — 5/5 (representative paths + sweep asserting no path starts with `/api/v1`)
- `evo-ai-frontend-community: pnpm exec tsc -b --noEmit` — clean
- `evo-ai-frontend-community: pnpm exec eslint` — no new findings (7 pre-existing `extractData<any>` = develop baseline)
- AC3 sweep: no other `apiAuth` consumer passes `/api/v1` literals (flagged matches were comments / another client's baseURL)
- Manual smoke (`/settings/users` → add user → populated role dropdown) pending environment — proxy :3030 was down during the fix; covered by the request-path spec meanwhile

## Changed Files

- `src/services/rbac/rbacService.ts`
- `src/services/rbac/rbacService.spec.ts` (new)

## Linked Issue

- EVO-1688 (related: EVO-1040)

## Summary by Sourcery

Fix RBAC API client paths to align with the apiAuth base URL and prevent doubled /api/v1 prefixes that caused RBAC requests to 404.

Bug Fixes:
- Correct RBAC service endpoints to use paths without the redundant /api/v1 prefix so role- and permission-related requests reach the auth service successfully.

Tests:
- Add vitest coverage to ensure rbacService calls use unprefixed paths and guard against reintroducing /api/v1-prefixed routes.

### DIFF
--- a/src/services/rbac/rbacService.spec.ts
+++ b/src/services/rbac/rbacService.spec.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+
+vi.mock('@/services/core', () => ({
+  apiAuth: {
+    get: vi.fn().mockResolvedValue({ data: { data: [] } }),
+    post: vi.fn().mockResolvedValue({ data: { data: {} } }),
+    put: vi.fn().mockResolvedValue({ data: { data: {} } }),
+    delete: vi.fn().mockResolvedValue({ data: {} }),
+  },
+}));
+
+import { apiAuth } from '@/services/core';
+import {
+  fetchRoles,
+  fetchRoleActions,
+  fetchRolePermissions,
+  createRole,
+  deleteRolePermission,
+} from './rbacService';
+
+const mockedGet = apiAuth.get as ReturnType<typeof vi.fn>;
+const mockedPost = apiAuth.post as ReturnType<typeof vi.fn>;
+const mockedDelete = apiAuth.delete as ReturnType<typeof vi.fn>;
+
+// EVO-1688 regression guard: apiAuth's baseURL already ends in /api/v1, so the
+// paths passed here must NOT repeat the prefix — doubled prefixes 404 on the
+// auth service (broke the user-creation role dropdown).
+describe('rbacService request paths (EVO-1688)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('fetchRoles targets /roles without the /api/v1 prefix', async () => {
+    await fetchRoles();
+    expect(mockedGet.mock.calls[0][0]).toMatch(/^\/roles\?/);
+  });
+
+  it('fetchRoleActions targets /resource_actions without the prefix', async () => {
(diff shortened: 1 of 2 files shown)


dpaes (approved):
## ✅ Approved — EVO-1688

Reviewed adversarially against the ACs; verified the bug premise and the fix end-to-end (cloned the PR head).

**AC verification**
- **AC1** ✓ — all 11 `rbacService` call sites de-prefixed; `grep -c '/api/v1' rbacService.ts` → 0. The new `rbacService.spec.ts` asserts the actual request-path argument (and sweeps that no path starts with `/api/v1`) — it does not mock away what it asserts.
- **AC2** ✓ (code-path) — `UserFormModal → useRoles()` (defaults `loadFull=false`) → the fixed `fetchRoles` `/roles?...`. The path that fed the empty dropdown is the one repaired. Runtime smoke deferred (proxy :3030 was down) — legitimate, scoped.
- **AC3** ✓ — swept all 14 `apiAuth` importers; no `apiAuth` call passes a `/api/v1` literal anywhere.

The "known limitation" (5 functions hitting auth routes that may not exist) is accurate: they're only reachable from the zero-importer `hooks/rbac/usePermissions.ts` (dead code); the live Admin Roles UI uses `rolesService`. No regression introduced.

**Non-blocking (no card needed, dev's discretion):** run the manual smoke on `/settings/users` when an env is up; CI is Sourcery-only so the vitest/tsc/eslint green is self-attested (spec inspected as correct); the 5 dead-route functions + dead `usePermissions` hook are a latent cleanup.

Squash-merging to `develop`.

PR #151, opened 2026-06-10
fix(frontend): smooth conversation-list scroll pagination (EVO-1672)

nickoliveira23:
## Summary

Follow-up to EVO-1617/EVO-1671: load-more on the conversation list was janky — fast scroll outran the prefetch (blank gap + pop-in), and visual testing during this fix surfaced two worse pre-existing behaviors that the wider prefetch made constant:

1. **Prefetch lookahead widened** — `PREFETCH_VIEWPORT_FACTOR` 1.5 → 2.5, `MIN_PREFETCH_DISTANCE_PX` 600 → 1000. Requests stay sequential (existing `loadingMoreRef` guard).
2. **Proportional loading cushion** — bottom placeholder 1 → 5 skeleton rows; slow networks read as a deliberate "loading more" affordance. (Removing the skeleton stays out of scope, per the card.)
3. **Full-list skeleton only on the empty/initial load** — `loadMoreConversations` flips the shared `conversationsLoading` flag, and the sidebar swapped the WHOLE list for 8 skeleton rows mid-scroll: list "vanished", scroll lost, flicker on return. With items on screen the list now stays mounted.
4. **Dropped the forced scrollTop restore after load-more** (scroll + button paths) — it compensated for the now-gone remount and, on slow connections, yanked the user back to wherever the trigger had fired if they kept scrolling during the request.

`DEFAULT_PAGE_SIZE` intentionally untouched (card's optional item) — the above closed the ACs without heavier payloads.

## Validation

- `pnpm exec vitest run .../ChatSidebar.spec.tsx` — **16 passing** (CA-1b re-pinned to the widened threshold; new cushion test count=5; new guard test: populated list + loading flag keeps the list visible)
- `pnpm exec eslint <changed>` — clean
- **Manual smoke (Slow 3G, multi-page account, by the reporter):** no blank gap on fast scroll; continuous cushion; the loaded list never vanishes; scroll position holds while pages load

## Changed Files

- `src/components/chat/chat-sidebar/ChatSidebar.tsx`
- `src/components/chat/chat-sidebar/ChatSidebar.spec.tsx`

## Linked Issue

- EVO-1672 (related: EVO-1617, EVO-1671, EVO-1106)

## Summary by Sourcery

Improve chat sidebar conversation list scroll pagination to avoid janky gaps and scroll jumps during incremental loading.

Bug Fixes:
- Prevent the conversation list from being replaced by a full-list skeleton during load-more when items are already visible, preserving scroll position.
- Remove forced scroll position restoration after loading more conversations so users are not pulled back while scrolling on slow connections.

Enhancements:
- Increase scroll prefetch distance and minimum prefetch threshold so additional conversation pages are requested earlier during fast scrolling or on slow networks.
- Show a multi-row skeleton cushion at the bottom of the list during load-more instead of a single row to provide clearer loading feedback.

Tests:
- Extend chat sidebar pagination tests to cover the widened prefetch threshold, verify that the populated list stays visible while loading, and assert the new multi-row loading cushion behavior.

### DIFF
--- a/src/components/chat/chat-sidebar/ChatSidebar.spec.tsx
+++ b/src/components/chat/chat-sidebar/ChatSidebar.spec.tsx
@@ -59,7 +59,9 @@ vi.mock('@/utils/chat/mediaLabels', () => ({
 }));
 
 vi.mock('../loading-states', () => ({
-  ConversationSkeleton: () => <div data-testid="skeleton" />,
+  ConversationSkeleton: ({ count }: { count?: number }) => (
+    <div data-testid="skeleton" data-count={count} />
+  ),
 }));
 
 vi.mock('../empty-states', () => ({
@@ -498,9 +500,10 @@ describe('ChatSidebar scroll pagination (EVO-1407)', () => {
     await screen.findByText('Test Contact');
 
     const scrollEl = document.querySelector('[data-tour="chat-conversations-list"]')!;
-    // clientHeight 600 → threshold = max(600, 900) = 900px. distanceToBottom = 700px:
-    // far beyond the old 120px gate, yet still triggers the prefetch.
-    setScrollDimensions(scrollEl, 5000, 600, 3700);
+    // EVO-1672: clientHeight 600 → threshold = max(1000, 1500) = 1500px.
+    // distanceToBottom = 1400px — beyond the previous 900px threshold, yet
+    // still triggers, pinning the widened lookahead.
+    setScrollDimensions(scrollEl, 5000, 600, 3000);
 
     await act(async () => { fireEvent.scroll(scrollEl); });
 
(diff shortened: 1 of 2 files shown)


dpaes (approved):
Approving — EVO-1672.

Surgical change (2 files, +57/-29). Acceptance criteria verified:
- Prefetch widened (`PREFETCH_VIEWPORT_FACTOR` 1.5→2.5, `MIN_PREFETCH_DISTANCE_PX` 600→1000), with a test pinning the trigger at 1400px.
- Bottom cushion 1→5 skeleton rows.
- Full-list skeleton now gated to the empty/initial load (the list no longer vanishes mid-scroll); the forced scroll-restore was removed.
- Load-more stays sequential (`loadingMoreRef` + context guard); APPEND merge, no full-list refetch.

EVO-1106 regression risk (scroll jumping to top) evaluated and ruled out: load-more APPENDs (preserves items/order, no remount) and appending below the viewport doesn't move `scrollTop`; no residual `scrollTop` writes remain.

Non-blocking nit: the 5-row cushion is fixed, not proportional to viewport height — fine vs the previous single row.

CI caveat: frontend doesn't run vitest/tsc in CI, but I ran them locally during review — vitest 16/16 and tsc clean. AC#1 ("no visible gap") is visual/runtime, validated by smoke only.

PR #140, opened 2026-06-05
feat(frontend): commit Journey node-type manifest as cross-repo parity source (EVO-1634)

nickoliveira23:
## Summary
Commits the **Journey node-type manifest** as the cross-repo source of truth for palette ↔ executor parity — part of **EVO-1634**.

The evo-flow coverage guard (`journey-execution.coverage.spec.ts`) reads this manifest to assert that **every** action node the Journey palette can produce has a non-`default` executor branch, so a newly-added palette node can no longer silently no-op at runtime.

- `journey-node-manifest.json` — committed list of the 25 palette `nodeType`s.
- `journey-node-manifest.spec.ts` — vitest that fails if the manifest drifts from the node types actually referenced in `JourneyFlowEditor.tsx`.

This is additive-only (no UI/behavior change); it locks the palette as the canonical list the runtime is checked against.

## Validation
- `pnpm exec tsc -b --noEmit`
- `pnpm test src/pages/Customer/Journey/journey-node-manifest.spec.ts` (manifest ↔ editor in sync)

## Changed Files
- `src/pages/Customer/Journey/journey-node-manifest.json`
- `src/pages/Customer/Journey/journey-node-manifest.spec.ts`

## Related PRs
- evo-flow#25 — executor node wiring + coverage guard that consumes this manifest
- evo-ai-crm-community#116 — CRM endpoints for the wired nodes

## Linked Issue
- EVO-1634

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Add a committed Journey node-type manifest and associated tests to enforce palette-to-executor parity across repositories.

New Features:
- Introduce a version-controlled Journey node-type manifest as the canonical list of palette node types.

Enhancements:
- Wire the Journey execution coverage guard to consume the node-type manifest as the source of truth for palette-to-executor parity.

Tests:
- Add a unit test to ensure the Journey node-type manifest stays in sync with the node types referenced by the Journey flow editor.

PR #157, opened 2026-06-12
refactor(message-templates): repoint to dedicated flat endpoint (EVO-1716)

daniloleonecarneiro:
## Summary

Point the global and channel template services at the new flat `/api/v1/message_templates` endpoint (CRM PR #140).

- **Global service**: drops the `GLOBAL_PARAM` / placeholder-inbox-id workaround — now a plain flat call.
- **Channel service**: CRUD uses the `inbox_id` query filter; **Meta sync calls unchanged**.
- **No UI/UX change** to the Settings template flow — internal wiring only.

## Test plan

- `globalMessageTemplatesService.spec.ts` + new `channels/__tests__/messageTemplatesService.spec.ts` + `MessageTemplates.spec.tsx`: **14/14 green**.
- Lint + typecheck clean.

## Notas

- Depends on CRM #140 (backend endpoint) and the auth-service RBAC PR being deployed.

### DIFF
--- a/src/pages/Customer/Settings/MessageTemplates/MessageTemplates.tsx
+++ b/src/pages/Customer/Settings/MessageTemplates/MessageTemplates.tsx
@@ -22,8 +22,6 @@ import {
 import { ChevronLeft, ChevronRight, Edit, LayoutTemplate, Plus, Search, Trash2 } from 'lucide-react';
 import { useLanguage } from '@/hooks/useLanguage';
 import { useUserPermissions } from '@/hooks/useUserPermissions';
-import { useAppDataStore } from '@/store/appDataStore';
-import EmptyState from '@/components/base/EmptyState';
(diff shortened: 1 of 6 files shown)


dpaes (approved):
## ✅ Approved — EVO-1716 (frontend: repoint to flat endpoint)

Verified the repoint is complete: both services (`globalMessageTemplatesService`, `channels/messageTemplatesService`) now hit `/message_templates` — global with no filter, channel-bound via `inbox_id`. The `GLOBAL_PARAM` / placeholder-inbox workaround and the now-dead no-inboxes empty state + `fetchInboxes` wiring are removed, and Meta sync stays on the inbox-scoped route. Services are repointed internally, so every caller follows automatically; no un-repointed caller of the old route remains in `src`. Tests updated accordingly.

**Deploy order:** must ship **last** — after the CRM cutover (#140) removes the old inbox-nested CRUD routes.

LGTM.

PR #158, opened 2026-06-12
docs(message-templates): drop stale channel_id filter mention (EVO-1716)

daniloleonecarneiro:
## Summary

Limpeza cosmética de follow-up. O backend removeu o filtro `channel_id` (não usado) do endpoint flat `/message_templates` (ver CRM #141), então os doc-comments do `globalMessageTemplatesService` referenciavam um param que não existe mais.

- Removidas as 2 menções a `inbox_id`/`channel_id` filter nos comentários.
- Mantido o `channel_id IS NULL` (descreve o estado da **coluna** dos templates channel-less, não um filtro de request).

Sem mudança de comportamento — só comentário.

## Test plan

- `eslint` no arquivo: limpo.
- `vitest` nos services de template (`globalMessageTemplatesService` + `channels/messageTemplatesService`): **11/11 green**.

## Notas

- Pareia com CRM #141 (remoção da superfície `channel_id` no backend). Pode mergear independente — é só doc.

## Summary by Sourcery

Documentation:
- Clarify that the flat /message_templates endpoint no longer accepts a channel_id filter and adjust comments to only reference inbox_id where applicable.

### DIFF
--- a/src/services/messageTemplates/globalMessageTemplatesService.ts
+++ b/src/services/messageTemplates/globalMessageTemplatesService.ts
@@ -7,8 +7,8 @@ import { extractData, extractResponse } from '@/utils/apiHelpers';
  * Global (channel-independent) message templates — EVO-1233 [6.4].
  *
  * These hit the flat, account-scoped `/message_templates` endpoint (EVO-1716).
- * With NO `inbox_id`/`channel_id` filter the backend lists/creates/updates/
- * deletes channel-less (`channel_id IS NULL`) templates. WhatsApp Cloud is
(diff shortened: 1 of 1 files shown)


dpaes (approved):
**Reviewed — approved ✅**

Doc-only follow-up paired with CRM #141. Verified against both message-template services: the global service (`globalMessageTemplatesService`) sends neither `inbox_id` nor `channel_id`, and the channel-bound service always sends `inbox_id` — so no consumer ever passed the `channel_id` query filter the backend dropped. The comment edits are accurate, and the retained `channel_id IS NULL` mention correctly describes the column state of channel-less templates, not a request filter. Zero behavior change.

PR #162, opened 2026-06-16
fix(frontend): preserve journey flow when editing metadata via the modal (EVO-1745)

nickoliveira23:
## Summary
`JourneyModal.handleSave` always built a synthetic **trigger-only** `flowData` (+ default `flowTriggers`) and sent it on **every** save — including updates. Editing an existing journey's name/description/active via the list modal therefore **wiped the flow built in the editor** (the backend replaces `flowData` wholesale on update). Silent data loss, no warning.

Fix (frontend only):
- **Update** now sends only `{ name, description, isActive }` → the backend keeps the stored `flowData`/`flowTriggers`/variables (`updateJourney` PATCHes the partial payload; `UpdateJourneyDto` is PartialType).
- **Create** is unchanged — a new journey still gets the default trigger-only `flowData` + `flowTriggers`.

## Security
None. Frontend payload shaping only; backend was already correct.

## Test plan
- `evo-ai-frontend-community: pnpm vitest run src/components/journey/JourneyModal.spec.tsx` → 2/2 (update sends no flowData/flowTriggers; create sends flowData)
- `pnpm exec tsc -b --noEmit` → clean for changed files (pre-existing unrelated error in TemplateParser.tsx:121)
- `eslint` (changed files) → clean
- **Live smoke:** built a real flow → edited the name via the list modal → reopened the editor → flow intact (was reduced to a lone trigger before).

## Changed Files
- `src/components/journey/JourneyModal.tsx`
- `src/components/journey/JourneyModal.spec.tsx` (new)

## Linked Issue
- EVO-1745

## Summary by Sourcery

Preserve existing journey flows when editing metadata in the modal by sending only metadata fields on update while keeping default flow initialization for creation.

Bug Fixes:
- Prevent existing journey flows from being overwritten with a default trigger-only flow when updating name, description, or active status via the modal.

Tests:
- Add unit tests for JourneyModal to verify that updates omit flowData/flowTriggers and creations include the default trigger-only flow.

### DIFF
--- a/src/components/journey/JourneyModal.spec.tsx
+++ b/src/components/journey/JourneyModal.spec.tsx
@@ -1,0 +1,87 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import JourneyModal from './JourneyModal';
+
+vi.mock('@/services', () => ({
+  journeyService: {
+    createJourney: vi.fn().mockResolvedValue({}),
+    updateJourney: vi.fn().mockResolvedValue({}),
+  },
+}));
+
+vi.mock('@/hooks/useLanguage', () => ({
+  useLanguage: () => ({
+    t: (key: string) => key,
+    currentLanguage: 'pt-BR',
+    changeLanguage: () => undefined,
+  }),
+}));
+
+vi.mock('sonner', () => ({
+  toast: { error: vi.fn(), success: vi.fn() },
+}));
+
+import { journeyService } from '@/services';
+
+beforeEach(() => vi.clearAllMocks());
+afterEach(() => vi.restoreAllMocks());
+
+const existingJourney = {
+  id: 'j1',
+  name: 'My Journey',
+  description: 'desc',
+  isActive: true,
+  flowData: {
+    nodes: [
+      { id: 'journey-trigger-node', type: 'journey-trigger-node' },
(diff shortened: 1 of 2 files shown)


dpaes (approved):
## ✅ Approved

The metadata-edit data-loss bug is fixed correctly. `JourneyModal.handleSave` now sends only `{ name, description, isActive }` on update; create still sends the synthetic trigger-only `flowData`/`flowTriggers`.

**Verified**
- The load-bearing backend contract holds: omitting `flowData`/`flowTriggers` truly preserves the stored flow. `journeys.service.update()` does `Object.assign(journey, dto) + save()` and `UpdateJourneyDto = PartialType(CreateJourneyDto)`; with the pinned `class-transformer@0.5.1` + `@nestjs/swagger` `PartialType`, omitted DTO props are **not** materialized as own-keys, so `Object.assign` never overwrites them (reproduced against the exact pinned versions).
- The update payload literally omits the keys (not just `undefined` values); the create path is byte-equivalent to before.
- The spec asserts the contract at the right layer (update omits `flowData`/`flowTriggers`; create includes the default trigger node first).

**Non-blocking (LOW)**
1. Clearing a description to empty via the modal is a no-op (`|| undefined`) — pre-existing, unchanged here.
2. No backend regression test guards the omit-preserves-flow contract — worth a future evo-flow test.

Caveat: frontend CI is Sourcery-only (vitest/tsc self-reported), but the live smoke was confirmed (build a flow → edit the name → flow intact). Merging to `develop`.

PR #164, opened 2026-06-17
fix(message-templates): stop infinite load loop and standardize grid (EVO-1759)

daniloleonecarneiro:
## Summary

Two UI fixes for the global Message Templates page (`Settings → Message Templates`), surfaced while testing the menu shipped in EVO-1233.

- **Infinite load loop (page unusable).** The list effect was `useEffect(() => loadTemplates(), [loadTemplates])`, and `loadTemplates` is a `useCallback` that depends on `can`. `useUserPermissions().can` is rebuilt inline on every render (no `useCallback`), so `loadTemplates` was never referentially stable → the effect re-ran every render → `setState` → re-render → an infinite fetch/loading loop. Now the effect gates on `permissionsReady` (`isReady`) and depends only on stable primitives (`page`, `debouncedSearch`), mirroring `CannedResponses`/`Labels`/`Teams`. This also removes a spurious "permission denied" toast that fired before the permission cache loaded.
- **Grid not on the standard components.** The table was hand-rolled (`Table`/`Card`/`Skeleton` + inline Edit/Delete buttons + a custom Prev/Next pager). Migrated to the shared **`BaseTable`** (`columns`/`actions` config, dropdown actions, internal loading + empty-state) + **`BasePagination`** (page numbers, "showing X–Y of Z", `sidebar` theme), matching the rest of Settings. Added `table.edit` / `table.delete` i18n keys across all 6 locales.

## Test plan

- `pnpm exec vitest run src/pages/Customer/Settings/MessageTemplates` — **4/4 green**, including a new regression test asserting `getTemplates` is called exactly once on mount (fails on the pre-fix effect, passes now).
- `pnpm exec tsc -b` — clean.
- `pnpm lint` (touched files) — clean.

## Notas

- No backend/API change; pure FE. The grid swap is behavior-preserving (same data, search, pagination).
- Builds on EVO-1233 (menu) + EVO-1716 (flat `/message_templates` endpoint), both already on develop.

## Linked Issue

- EVO-1759

## Summary by Sourcery

Fix loading behavior and modernize the Message Templates settings grid to align with shared UI patterns.

Bug Fixes:
- Prevent the Message Templates page from entering an infinite fetch/render loop by gating template loading on ready permissions and stable pagination/search state.

Enhancements:
- Replace the custom Message Templates table and pager with shared BaseTable and BasePagination components, including proper pagination metadata and standardized row actions.
- Add reusable edit/delete table action labels to the Message Templates i18n resources across all supported locales.

Tests:
- Add a regression test ensuring templates are fetched exactly once on mount to guard against future infinite render/fetch loops.
- Adjust pagination tests to account for the new BasePagination behavior and pagination metadata.

### DIFF
--- a/src/i18n/locales/en/messageTemplates.json
+++ b/src/i18n/locales/en/messageTemplates.json
@@ -10,7 +10,9 @@
     "category": "Category",
     "status": "Status",
     "language": "Language",
-    "actions": "Actions"
+    "actions": "Actions",
+    "edit": "Edit",
+    "delete": "Delete"
   },
   "status": {
     "approved": "Approved",
--- a/src/i18n/locales/es/messageTemplates.json
+++ b/src/i18n/locales/es/messageTemplates.json
@@ -10,7 +10,9 @@
     "category": "Categoría",
     "status": "Estado",
     "language": "Idioma",
-    "actions": "Acciones"
+    "actions": "Acciones",
+    "edit": "Editar",
+    "delete": "Eliminar"
   },
   "status": {
     "approved": "Aprobada",
--- a/src/i18n/locales/fr/messageTemplates.json
+++ b/src/i18n/locales/fr/messageTemplates.json
@@ -10,7 +10,9 @@
     "category": "Catégorie",
     "status": "Statut",
     "language": "Langue",
-    "actions": "Actions"
+    "actions": "Actions",
+    "edit": "Modifier",
+    "delete": "Supprimer"
   },
   "status": {
     "approved": "Approuvé",
--- a/src/i18n/locales/it/messageTemplates.json
+++ b/src/i18n/locales/it/messageTemplates.json
@@ -10,7 +10,9 @@
     "category": "Categoria",
     "status": "Stato",
     "language": "Lingua",
-    "actions": "Azioni"
+    "actions": "Azioni",
+    "edit": "Modifica",
+    "delete": "Elimina"
   },
   "status": {
     "approved": "Approvato",
--- a/src/i18n/locales/pt-BR/messageTemplates.json
+++ b/src/i18n/locales/pt-BR/messageTemplates.json
@@ -10,7 +10,9 @@
     "category": "Categoria",
     "status": "Status",
     "language": "Idioma",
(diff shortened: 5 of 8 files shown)


dpaes (approved):
**Approved** ✅ — reviewed with an adversarial multi-agent pass (loop-fix correctness, BaseTable/BasePagination contract parity, i18n, tests, security/scope).

**The two fixes hold up:**
- **Infinite load loop (High):** root cause confirmed — `useUserPermissions().can` is a fresh closure every render, so `loadTemplates` (a `useCallback` listing `can`) is never referentially stable, and `useEffect(..., [loadTemplates])` re-ran every render → infinite fetch. The fix (gate on `permissionsReady` + depend only on `[permissionsReady, page, debouncedSearch]`) is correct. Adding `page`/`debouncedSearch` to the deps — unlike the reference pages that use only `[permissionsReady]` — is justified here because this page paginates/searches **server-side**. The regression test is faithful (fails on the old dep list, passes now).
- **Grid migration (Medium):** `BaseTable` + `BasePagination` props and types are correct against the actual component signatures. Bonus: RBAC is now **stricter** — Edit/Delete are gated on `can('message_templates', …)` instead of always rendering.

**i18n:** all 6 locales add `table.edit/table.delete` (properly translated); every `t()` key used on the page resolves.

**Non-blocking nits (follow-up, not gating this merge):**
- **[low]** `isLoading` is initialized `true`; if `permissionsReady` never becomes true (permissions-fetch error or an account with zero account-permissions), `loadTemplates` never runs and the table is stuck on the loading spinner forever. The reference pages (`CannedResponses`/`Labels`) initialize `loading=false` and fall through to the empty state. One-line fix: `useState(false)` for `isLoading`.
- **[low/nit]** the pagination test has a small debounce-mount-timer flake window and uses a real `setTimeout(50ms)` instead of fake timers; the removed hand-rolled pager left `pagination.previous/next/pageOf` as dead i18n keys.

Note: this repo's CI runs Sourcery only (no vitest/tsc/lint), so the green test/tsc claims are dev-reported — worth a local `pnpm test` before relying on them.

PR #152, opened 2026-06-10
feat(frontend): channel hub shows real live connectivity per inbox (EVO-1674)

nickoliveira23:
## Summary
- Replaces the hub's DERIVED status (EVO-1554 / PR #145) with the REAL state from `/inboxes` (`connection_state` / `health_source` / `last_sync` — backend PR below): per-inbox connected/disconnected/testing/error, relative last-connectivity time, and an explicit 'not monitored' label for channel types without health support.
- New `useLiveChannelStatus` hook: one connection-state probe per evolution WhatsApp inbox through the existing CRM proxy (`GET /evolution/instances?instanceName=...`, credentials resolved server-side), overlaying the stored state with per-inbox loading (pulsing dot) and failure (tooltip) handling. Other providers/types keep the stored state — explicit degrade.
- Type badge and summary gain a red `error` level (any disconnected/error inbox); aggregation: error > attention (pending) > active; unknown counts as active (unmonitored ≠ broken).
- i18n: new status keys in all six locales (pt-BR, en, es, fr, it, pt), following the EVO-1554 pattern.
- Legacy fallback kept for payloads predating the backend field (reauthorization_required → error).

## Security
- No new endpoints; probes reuse the authenticated CRM proxy. No secrets rendered.

## Test plan
- `pnpm exec tsc -b --noEmit` — clean
- `pnpm exec vitest run src/utils/channelStatus.spec.ts src/components/channels/ChannelTypeHub.spec.tsx src/hooks/channels/useLiveChannelStatus.spec.ts src/i18n/locales/i18n-parity.spec.ts` — 188 tests green (state mapping, overlay precedence, hook loading/failure paths, hub rendering, locale parity)
- ESLint clean on touched files

## Changed Files
- `src/utils/channelStatus.ts` (+ spec) — real-state mapping, overlay, error aggregation, formatLastSync
- `src/hooks/channels/useLiveChannelStatus.ts` (+ spec) — live evolution probe
- `src/components/channels/ChannelTypeCard.tsx`, `ChannelTypeHub.tsx` (+ spec), `ChannelStatusBadge.tsx`
- `src/types/channels/inbox.ts` — new Inbox fields
- `src/i18n/locales/{pt-BR,en,es,fr,it,pt}/channels.json`

## Related PRs
- evo-ai-crm-community: https://github.com/evolution-foundation/evo-ai-crm-community/pull/133 (exposes the /inboxes fields — merge first or together)

## Linked Issue
- EVO-1674

PR #165, opened 2026-06-17
feat(channels): pick a global template for inbox greeting/out-of-office (EVO-1760)

daniloleonecarneiro:
## Summary

First UI gap from the EVO-1760 audit: global (channel-less) message templates were already resolvable by the backend send paths, but no screen let a user attach one. This wires the **inbox greeting / out-of-office** path end-to-end.

- New reusable `GlobalTemplateSelect` — lists the global catalog via `globalMessageTemplatesService` (flat `/message_templates`, no `inbox_id`), fetched once on mount.
- `GreetingSettingsForm` + `BusinessHoursForm`: a **"use global template"** toggle swaps the free-text field for the picker, writing `greeting_message_template_id` / `out_of_office_message_template_id` (clearing it returns to free text — the backend prefers the template when the id is set).
- `ChannelSettings` wires both ids into the inbox update payload; `Inbox` type + 6 locales updated.

Depends on the CRM PR (permit + serialize the two ids).

## Test plan

- `GreetingSettingsForm.spec.tsx` — **4/4** (free-text default; starts in template mode when an id is set; toggle on shows the picker; toggle off clears the ref).
- `GlobalTemplateSelect.spec.tsx` — **2/2** (lists globals from the flat endpoint + emits the id; disabled when none).
- `BusinessHoursForm.spec.tsx` — added an out-of-office template-mode test. Note: this spec uses `importActual('@evoapi/design-system')` and does not complete in my local sandbox (pre-existing on develop, not introduced here); it validates in CI.
- `tsc -b` clean; eslint clean on touched files.

## Notas

- WhatsApp Cloud is intentionally out of scope (channel-bound; never in the global catalog).
- Remaining EVO-1760 items (automation dropdown, chat picker, bot auto-reply) stay open on the card.

## Linked issue
- EVO-1760

## Summary by Sourcery

Add support for selecting global message templates for inbox greeting and out-of-office messages and wire the chosen template IDs through channel settings and inbox updates.

New Features:
- Introduce a reusable GlobalTemplateSelect component that lists global (channel-less) message templates from the shared catalog.
- Allow greeting messages to use either free text or a selected global message template in channel settings.
- Allow out-of-office messages in business hours settings to use either free text or a selected global message template, persisting the template ID with the inbox.

Enhancements:
- Extend inbox and channel settings types and payloads to include greeting and out-of-office message template IDs.
- Localize new greeting and business-hours template selection labels and help text across supported locales.

Tests:
- Add focused tests for GreetingSettingsForm to cover default free-text mode, template-initialized mode, toggle behavior, and clearing the template reference.
- Add coverage in BusinessHoursForm tests for initializing in template mode when an out-of-office template ID is present.
- Add unit tests for the GlobalTemplateSelect component to verify listing global templates, emitting selections, and handling an empty catalog.

### DIFF
--- a/src/components/channels/settings/BusinessHoursForm.spec.tsx
+++ b/src/components/channels/settings/BusinessHoursForm.spec.tsx
@@ -60,6 +60,12 @@ vi.mock('@evoapi/design-system', async () => {
   };
 });
 
+vi.mock('./GlobalTemplateSelect', () => ({
+  default: ({ value }: { value?: string | null }) => (
+    <div data-testid="global-template-select">{value || 'none'}</div>
+  ),
+}));
+
 describe('BusinessHoursForm', () => {
   const mockOnUpdate = vi.fn().mockResolvedValue(undefined);
   const defaultProps = {
@@ -90,9 +96,22 @@ describe('BusinessHoursForm', () => {
   describe('Button disabled state', () => {
     it('should not disable button when business hours is disabled even if there are validation errors', () => {
       render(<BusinessHoursForm {...defaultProps} workingHoursEnabled={false} />);
-      
+
       const updateButton = screen.getByRole('button', { name: /settings\.businessHours\.buttons\.update/i });
       expect(updateButton).not.toBeDisabled();
+    });
+  });
+
+  describe('Out-of-office template', () => {
+    it('starts in template mode when an out-of-office template id is set', () => {
+      render(
+        <BusinessHoursForm
+          {...defaultProps}
+          workingHoursEnabled={true}
+          outOfOfficeMessageTemplateId="tpl-9"
+        />,
+      );
+      expect(screen.getByTestId('global-template-select')).toHaveTextContent('tpl-9');
     });
   });
 });
--- a/src/components/channels/settings/BusinessHoursForm.tsx
+++ b/src/components/channels/settings/BusinessHoursForm.tsx
@@ -16,6 +16,7 @@ import { toast } from 'sonner';
 import { useLanguage } from '@/hooks/useLanguage';
 
 import BusinessDay from './BusinessDay';
+import GlobalTemplateSelect from './GlobalTemplateSelect';
 import {
   TimeSlot,
   TimeZone,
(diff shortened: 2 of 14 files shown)


dpaes (approved):
**Approved** ✅ — reviewed with an adversarial multi-agent pass (backend contract, FE wiring/state, **end-to-end closure**, i18n, tests, scope/security/tenancy). This is the FE half of EVO-1760 gap #1 (greeting/out-of-office); paired with CRM #152.

**End-to-end is real, not an inert toggle:** verified the columns this exposes are consumed by a live send path — `HookExecutionService#trigger_templates` → `Template::Greeting`/`OutOfOffice` gate on `inbox.*_message_template_id.present?`, content resolves `template_content_for(template_id) || inline` (template preferred), via the global-aware `SendResolver` with a safe fallback to free text on miss/render error.

**FE wiring is correct:** `GlobalTemplateSelect` fetches once on mount, is unmount-guarded, and is disabled on empty/loading; `ChannelSettings` threads both ids through initial state, load mapping, and the save payloads; `Inbox` type updated. i18n is complete across all 6 locales (both `businessHours` and `greeting` sections), properly translated.

**Non-blocking nits (follow-up):**
- **[medium, UX]** When "use template" is ON but no template is selected (or the catalog is empty → the picker is rendered disabled, a dead end), Save still persists the stale free-text + a `null` ref with no feedback. The backend falls back to free text safely, so nothing breaks — but the user's intent is silently dropped. Suggest disabling Save / showing a "select a template" hint, and surfacing the empty-catalog case.
- **[low]** `useTemplate` only ever flips ON (never back OFF) on a props-driven clear; narrow window, mirrors the existing pattern.
- **[low]** `GlobalTemplateSelect` fetches `per_page: 200` with no search/pagination → templates beyond 200 are unselectable. (`per_page: -1` is explicitly forbidden by the service, so 200 is the correct choice; consider a searchable variant before reusing this for the deferred automation/chat pickers.)

Note: FE CI runs Sourcery only (no vitest/tsc/lint) — please confirm `vitest` locally before relying on the green claim.

PR #166, opened 2026-06-17
fix(frontend): retain label/segment trigger action on save (EVO-1754)

nickoliveira23:
## Summary
The Label/Segment journey trigger action `<Select>` received a `|| default` **display** value (`labelAction || 'applied'`, `segmentAction || 'entered'`) that was never committed to `formData`. Leaving the dropdown untouched (its default already shows "Applied"/"Entered") saved `undefined`. The canvas card then used a **different** default for `undefined`, so a node configured as "applied" rendered as **"removed"** (and segment "entered" rendered as "exits").

- `JourneyTriggerPanel.tsx` — `handleSave` persists the effective default: `labelAction ?? 'applied'`, `segmentAction ?? 'entered'`, so the saved value matches the dropdown.
- `JourneyTriggerNode.tsx` — the canvas card aligns its default with the editor/runtime (treat `undefined` as applied/entered via `!== 'removed'` / `!== 'exited'`), which also corrects legacy nodes already saved with `undefined`.

## Behavioral note
Existing journeys saved with no action (`undefined`) now render "applied"/"enters" on the canvas instead of "removed"/"exits". This is a **correction**: the runtime already treated unset as applied/added, and the editor input already defaulted to applied/entered — the canvas card was the only consumer disagreeing.

## Sweep
The card mentioned a possible "Wait node labelAction" — it does not exist in code. The only sibling sharing this exact pattern with a visible canvas mismatch is `segmentAction`, fixed here too. `customAttributeOperator` / `webhookMethod` share the `|| default` input pattern but their canvas rendering already defaults consistently → no visible bug, left out of scope.

## Out of scope — follow-up filed
Code review found that the **evo-flow runtime ignores the action entirely**: `label.trigger.ts` matches both `*.added` and `*.removed` (and `segment.trigger.ts` both `entered`/`exited`) without filtering by the configured action — so a trigger set to "applied" still fires on "removed". That is the backend half and is tracked in **EVO-1763** (High).

## Validation
- `evo-ai-frontend-community: pnpm exec tsc -b --noEmit` → clean
- `evo-ai-frontend-community: pnpm exec eslint <changed files>` → clean
- `evo-ai-frontend-community: pnpm vitest run JourneyTriggerPanel.spec.tsx JourneyTriggerNode.spec.tsx` → 17/17
- Live smoke (confirmed by Nickolas): Label trigger left at default → canvas shows "applied" (was "removed"). Segment path is identical root cause + unit-covered (segments backend/evo-flow not running locally).

## Changed Files
- `src/components/journey/nodes/trigger/JourneyTriggerPanel.tsx`
- `src/components/journey/nodes/trigger/JourneyTriggerNode.tsx`
- `src/components/journey/nodes/trigger/JourneyTriggerPanel.spec.tsx`
- `src/components/journey/nodes/trigger/JourneyTriggerNode.spec.tsx` (new)

## Linked Issue
- EVO-1754

## Summary by Sourcery

Align label and segment trigger actions between the editor form and canvas card so that default actions are correctly persisted and displayed.

Bug Fixes:
- Ensure label triggers default to and persist the "applied" action when the action is left unchanged before saving.
- Ensure segment triggers default to and persist the "entered" action when the action is left unchanged before saving.
- Make the trigger canvas card treat missing label/segment actions as applied/entered instead of removed/exited, correcting existing nodes with unset actions.

Tests:
- Add unit tests to verify label and segment actions are saved with the correct defaults and preserve explicit selections in the trigger panel.
- Add unit tests to verify the trigger canvas card renders correct label/segment action text when actions are unset or explicitly configured.
- Stub label and segment services in trigger panel tests to avoid network calls during test runs.

### DIFF
--- a/src/components/journey/nodes/trigger/JourneyTriggerNode.spec.tsx
+++ b/src/components/journey/nodes/trigger/JourneyTriggerNode.spec.tsx
@@ -1,0 +1,54 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { ReactFlowProvider } from '@xyflow/react';
+import { JourneyTriggerNode, type JourneyTriggerNodeData } from './JourneyTriggerNode';
+import '@/i18n/config';
+
+// BaseFlowNode renders @xyflow/react Handles, so the node needs a flow context.
+function renderNode(data: Partial<JourneyTriggerNodeData>) {
+  render(
+    <ReactFlowProvider>
+      <JourneyTriggerNode
+        id="node-1"
+        selected={false}
+        data={{ label: 'Trigger', triggerType: 'manual', ...data } as JourneyTriggerNodeData}
+      />
+    </ReactFlowProvider>,
+  );
+}
+
+// EVO-1754: the canvas card must default an unset action to the SAME value the
+// editor + runtime assume (label → applied, segment → entered), not the opposite.
+describe('JourneyTriggerNode — action defaults on the canvas card (EVO-1754)', () => {
+  it('renders a Label trigger with no labelAction as "applied" (not "removed")', () => {
+    renderNode({ triggerType: 'label', labelId: 'l1', labelName: 'vip' });
+    expect(screen.getByText(/"vip"\s+applied/i)).toBeTruthy();
+    expect(screen.queryByText(/removed/i)).toBeNull();
+  });
+
+  it('renders an explicit labelAction "removed" as "removed"', () => {
+    renderNode({
+      triggerType: 'label',
+      labelId: 'l1',
+      labelName: 'vip',
+      labelAction: 'removed',
+    });
+    expect(screen.getByText(/"vip"\s+removed/i)).toBeTruthy();
+  });
+
+  it('renders a Segment trigger with no segmentAction as "enters" (not "exits")', () => {
+    renderNode({ triggerType: 'segment', segmentId: 's1', segmentName: 'leads' });
+    expect(screen.getByText(/enters\s+"leads"/i)).toBeTruthy();
+    expect(screen.queryByText(/exits/i)).toBeNull();
+  });
+
+  it('renders an explicit segmentAction "exited" as "exits"', () => {
+    renderNode({
+      triggerType: 'segment',
+      segmentId: 's1',
+      segmentName: 'leads',
+      segmentAction: 'exited',
+    });
+    expect(screen.getByText(/exits\s+"leads"/i)).toBeTruthy();
+  });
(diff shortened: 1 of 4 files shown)


dpaes (approved):
**Approved** ✅ — reviewed with an adversarial multi-agent pass (default-alignment correctness, sweep completeness, test validity).

**The reported bug is fixed correctly.** Root cause confirmed: the Label/Segment action `<Select>` defaulted to applied/entered for **display only** (`labelAction || 'applied'`), never committed to `formData`, so an untouched dropdown saved `undefined` — and the canvas card used the **opposite** default for `undefined`, rendering "applied" as "removed". The fix makes all three FE consumers agree (undefined → applied/entered): editor input, `handleSave` (`?? 'applied'` / `?? 'entered'`), and the canvas card (`!== 'removed'` / `!== 'exited'`). The type union is exactly `'applied'|'removed'` / `'entered'|'exited'` + undefined, so `!==` can't misclassify any valid value; legacy nodes saved with `undefined` are corrected too, consistent with the runtime's unset-default (EVO-1763). Regression tests genuinely fail on the pre-fix code and pass now. Filing the runtime matcher half as **EVO-1763** (High, code-traced) is exactly the right call.

**Non-blocking follow-ups (do not gate this PR — the reported bug is fixed):**
- **[low] The sweep missed two siblings, and one scoping claim is inaccurate.** The PR states *"Wait node labelAction does not exist in code"* — it does: `WaitPanel.handleSave` (WaitNode.tsx:22,24,202 + WaitEventConfig/WaitConditionConfig) has the **same** persist-`undefined` pattern, unfixed; and `CampaignTriggerConfig`/`NewCampaign` ships `undefined` into `trigger_config` the same way. Both are **latent**, not visible: their canvas/summary either omit the action or already default consistently, and `undefined` is harmless at runtime today (evo-flow ignores/defaults the action per EVO-1763). Worth folding the same `?? default` into `WaitPanel.handleSave` and `CampaignTriggerConfig` in a follow-up.
- **[low] `JourneyFlowEditor.tsx` (the `flowTriggers` metadata copy sent to evo-flow) drops `segmentAction`/`segmentName` and passes `labelAction` raw.** Pre-existing, harmless now (the real persistence path is `flowData.nodes`, which round-trips correctly). **Heads-up for the EVO-1763 implementer:** if the runtime starts honoring the action via `flowTriggers.metadata`, `segmentAction` would be silently dropped here.

Note: FE CI runs Sourcery only (no vitest/tsc/eslint), so the green test/tsc claims are dev-reported (verified by inspection here).

PR #77, opened 2026-05-16
feat(EVO-1374): render carousel cards in chat MessageBubble

pastoriniMatheus:
## Summary

- Add `MessageCarousel` component for rendering `content_type: cards` as horizontal scrollable cards
- Add `cards` case in `MessageBubble.tsx` switch to wire the new component
- Each card displays: image (from `media_url`), title, description, and action buttons

## Root cause

`MessageBubble` had no handler for `content_type == 'cards'`, so carousel messages fell through to the default text renderer. Operators saw plain text instead of visual cards.

## Changes

| File | What |
|------|------|
| `MessageCarousel.tsx` (new) | Renders items from `content_attributes.items` as snap-scroll horizontal cards with image, title, description, buttons. Chevron arrows for navigation. |
| `MessageBubble.tsx` | Import `MessageCarousel` + `case 'cards'` in content type switch |

## Design decisions

- **Snap scroll** with `snap-x snap-mandatory` for smooth card-by-card navigation
- **Semi-transparent styling** consistent with existing components (`bg-white/5`, `border-white/20`, `backdrop-blur-sm`)
- **Read-only** — buttons are displayed but not interactive (operator sees what was sent to customer)
- **Lazy loading** images with `loading="lazy"`
- **Responsive** — card width fixed at 220px, container scrolls horizontally

## Test checklist

- [x] `npx tsc --noEmit` — zero errors
- [x] Visual: carousel renders in CRM chat with cards, images, descriptions and buttons
- [x] Scroll arrows appear when multiple cards present
- [x] Dark mode consistent styling
- [x] Empty items gracefully falls back to text

## Known limitation

Carousel images sent via Evolution Go may not display on WhatsApp when `media_url` redirects (HTTP 301/302). This is an Evolution Go bug (EVO-1377), not a CRM issue. Direct image URLs work correctly.

## Related

- EVO-1373 (backend carousel — PR #74 on evo-ai-crm-community)
- EVO-1377 (Evolution Go image redirect bug)

## Summary by Sourcery

Add support for rendering carousel card messages in the chat UI.

New Features:
- Introduce a MessageCarousel component to display card-based message content as a horizontally scrollable carousel with images, text, and actions.
- Wire the new cards content type into MessageBubble so messages with card payloads render as visual carousels instead of plain text.

PR #168, opened 2026-06-19
fix(frontend): unwrap success envelope in journeyService variable + journey reads (EVO-1836)

nickoliveira23:
## Summary
Part of **EVO-1836** (cross-repo — journey variables broken). `journeyService` read `response.data` expecting the array, but evo-flow returns the success envelope `{success, data, meta}`. Two user-visible symptoms:
- **Crash on create:** `updateJourneyVariables` returned the envelope object → `setVariables(<object>)` → `TypeError: variables.map is not a function` (VariableSelect.tsx:121).
- **Variables never display:** `getJourneyVariables` guarded `Array.isArray(response.data)` and fell to `[]` (the envelope is an object).

Fix: unwrap via the existing `extractData` helper (the rest of the service already uses it) in `getJourneyVariables`, `updateJourneyVariables`, and — same latent envelope bug surfaced in review — `getJourneysByTriggerType` and `duplicateJourney`.

## Validation
- `pnpm exec tsc -b --noEmit` → clean
- `npx vitest run journeyService.spec` → 9/9 (incl. 3 new variable-unwrap specs: array, crash-regression, [] fallback)
- `npx eslint` → no new errors (the file's pre-existing `no-explicit-any` debt left untouched)
- **Live:** create a journey variable → persists → ENV/trigger dropdown shows it; API `GET /journeys/:id/variables` returns the array.
- Adversarial multi-agent review: confirmed findings fixed; partial-save / race concerns refuted.

## Changed Files
- `src/services/journeys/journeyService.ts`
- `src/services/journeys/journeyService.spec.ts`

## Related PRs
- evo-flow (backend half: cache reconstruction preserves `variables` + refresh on write) — same EVO-1836.

## Linked Issue
- EVO-1836

## Summary by Sourcery

Unwrap journey-related API success envelopes so journey data and variables methods consistently return plain arrays or objects instead of the envelope.

Bug Fixes:
- Ensure getJourneyVariables returns the variables array (or an empty array) instead of the success envelope object to prevent frontend crashes and missing variables.
- Ensure updateJourneyVariables returns the updated variables array rather than the envelope payload.
- Ensure duplicateJourney correctly unwraps the response envelope to return a Journey object.
- Ensure getJourneysByTriggerType unwraps the response envelope and safely falls back to an empty array when the payload is not an array.

Tests:
- Add unit tests covering envelope unwrapping for journey variable reads and writes, including regression coverage for the previous crash and non-array payload fallback.

### DIFF
--- a/src/services/journeys/journeyService.spec.ts
+++ b/src/services/journeys/journeyService.spec.ts
@@ -94,3 +94,45 @@ describe('journeyService session methods — envelope unwrap', () => {
     expect(result.data.deleted).toBe(5);
   });
 });
+
+describe('journeyService variable methods — envelope unwrap (EVO-1836)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('getJourneyVariables unwraps the { success, data } envelope into the variables array', async () => {
+    const vars = [{ id: 'var_1', name: 'lead_score', type: 'number', defaultValue: '0' }];
+    vi.mocked(apiEvoFlow.get).mockResolvedValue({
+      data: { success: true, data: vars },
+    } as never);
+
+    const result = await journeyService.getJourneyVariables('journey-1');
+
+    expect(apiEvoFlow.get).toHaveBeenCalledWith('/journeys/journey-1/variables');
+    expect(Array.isArray(result.data)).toBe(true);
+    expect(result.data).toEqual(vars);
+  });
+
+  it('getJourneyVariables returns [] when the unwrapped payload is not an array', async () => {
+    vi.mocked(apiEvoFlow.get).mockResolvedValue({
+      data: { success: true, data: null },
+    } as never);
+
+    const result = await journeyService.getJourneyVariables('journey-1');
+
+    expect(result.data).toEqual([]);
+  });
+
+  it('updateJourneyVariables returns the variables array, not the envelope object (the crash regression)', async () => {
+    const vars = [{ id: 'var_1', name: 'lead_score', type: 'number', defaultValue: '0' }];
+    vi.mocked(apiEvoFlow.post).mockResolvedValue({
+      data: { success: true, data: vars, meta: { timestamp: 'x' } },
+    } as never);
+
+    const result = await journeyService.updateJourneyVariables('journey-1', vars);
+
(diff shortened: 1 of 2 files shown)


dpaes (approved):
**Review — EVO-1836 (approved ✅)**

Independent cross-repo verification (this PR #168 + evo-flow #77). Verdict: **correct and complete, no blockers.**

**Root cause confirmed.** evo-flow wraps every success in `{success,data,meta}` (global `ResponseTransformInterceptor`), but `journeyService` read `response.data` (the envelope). On create, `updateJourneyVariables` returned the envelope object → `setVariables(<object>)` → `variables.map is not a function` at `VariableSelect.tsx:121`. The `extractData` unwrap fixes it across all 4 methods; the `Array.isArray(...) ? data : []` guards cover null/non-array, and `extractData`'s fallback to `response.data` keeps non-enveloped responses working.

**Caller-impact (the two sibling shape changes): safe.**
- `duplicateJourney`: sole caller `Journey.tsx:111` ignores the return value.
- `getJourneysByTriggerType`: no live caller in the repo.
- `getJourneyVariables` / `updateJourneyVariables`: only `useJourneyVariables.ts` consumes them; all downstream `.map()` consumers (`VariableSelect`, `VariableInput`, `VariableTextarea`, `EnvironmentManager`) now always receive an array.

No call site relied on the old (buggy) envelope shape.

**Non-blocking (no action required to merge):**
- Coverage: the `duplicateJourney` and `getJourneysByTriggerType` unwrap changes are untested; the `data:null` test is a false-green (passes both before and after the fix). The two load-bearing variable tests are genuine (fail on revert).
- `useJourneyVariables.ts:60` calls `setVariables(response.data)` without `|| []` (safe only because the service now guarantees an array) — cosmetic, asymmetric with line 38.
- Out-of-scope sibling bug worth a follow-up: `campaignsService.ts` (same `apiEvoFlow` envelope) still returns the raw envelope at `getCampaigns`/`getCampaignStats`/`bulkAction` (l.36/97/103) and hand-rolls `response.data.data` without the `extractData` fallback elsewhere — the same bug class this PR standardized away in journeys.

Approving + squash-merging.

PR #172, opened 2026-06-22
feat(frontend): custom attribute action in automation builder (EVO-1751)

nickoliveira23:
## Summary
- New **"Update custom attribute"** action in the simple automation rule builder (`ActionRow`): pick a contact/conversation/pipeline-item custom attribute and set its value with a **type-aware input** (text / number / date / datetime / link / list / checkbox).
- Custom attribute definitions are loaded into the automation form data (`useAutomationFormData`); the action is registered in the Zod action union, the `AutomationActionType` type, and the i18n action list across **all 6 locales**.
- New `CustomAttributeValueInput` renders the value editor by `attribute_display_type`. The attribute select stores `attribute_key` + `attribute_model` (composite), matching the backend contract.
- Only the 3 backend-supported models are offered (contact / conversation / pipeline_item).
- `ConditionRow`: narrowed the option-loader key type to exclude `customAttributes` so the new form-data field doesn't widen the condition option union (type fix).

## Security
- No `dangerouslySetInnerHTML`; value handled as a string; reuses the existing `customAttributesService` / `GET /custom_attribute_definitions` endpoint.

## Test plan
- `frontend: pnpm exec tsc -b --noEmit` (clean for changed files; the only remaining errors are 3 pre-existing `StageAutomationRules.tsx` ones, unrelated)
- `frontend: pnpm test -- src/components/automation/CustomAttributeValueInput.spec.tsx src/components/automation/ActionRow.spec.tsx` (10 passing)
- `frontend: pnpm lint` on the changed files (clean)
- Manual smoke: action renders the attribute select (with the model suffix) + a list-type value dropdown; rule saved and fired on `conversation_updated`.

## Changed Files
- `src/components/automation/ActionRow.tsx`
- `src/components/automation/CustomAttributeValueInput.tsx` (new) + `.spec.tsx` (new)
- `src/components/automation/ConditionRow.tsx`
- `src/hooks/automation/useAutomationFormData.ts`
- `src/pages/Customer/Automation/registries/actionRegistry.ts`, `registries/index.ts`
- `src/services/automation/automationService.ts`
- `src/types/automation/automation.ts`
- `src/i18n/locales/{en,pt-BR,es,fr,it,pt}/automation.json`

## QA Notes
- Code-review findings addressed: `ActionRow.spec` `emptyFormData` kept type-complete (`customAttributes: []`); guarded the `model::key` parse against a missing separator.

## Related PRs
- crm (backend handler): https://github.com/evolution-foundation/evo-ai-crm-community/pull/158

## Linked Issue
- EVO-1751

### DIFF
--- a/src/components/automation/ActionRow.spec.tsx
+++ b/src/components/automation/ActionRow.spec.tsx
@@ -21,6 +21,7 @@ const emptyFormData: AutomationFormData = {
   messageTypes: [],
   cannedResponses: [],
   messageTemplates: [],
+  customAttributes: [],
 };
 
 function Wrapper({
--- a/src/components/automation/ActionRow.tsx
+++ b/src/components/automation/ActionRow.tsx
@@ -591,6 +591,7 @@ function CustomAttributeParam({
 
         const pickAttribute = (value: string) => {
           const sep = value.indexOf('::');
+          if (sep === -1) return;
           const model = value.slice(0, sep);
           const key = value.slice(sep + 2);
           field.onChange([{ custom_attribute_key: key, custom_attribute_model: model, custom_attribute_value: '' }]);
--- a/src/components/automation/ActionRow.tsx
+++ b/src/components/automation/ActionRow.tsx
@@ -20,6 +20,16 @@ import {
 import type { AutomationFormData } from '@/hooks/automation/useAutomationFormData';
 import type { AutomationActionType } from '@/types/automation';
 import type { MessageTemplateVariable } from '@/hooks/automation/useAutomationFormData';
+import CustomAttributeValueInput from './CustomAttributeValueInput';
+
+// Custom attribute models that map to a concrete record in automation scope and
+// can therefore be written by the update_custom_attribute action (EVO-1751).
+// pipeline / pipeline_stage attributes are config-level (no per-run record).
+const UPDATABLE_ATTRIBUTE_MODELS: string[] = [
+  'conversation_attribute',
+  'contact_attribute',
+  'pipeline_item_attribute',
+];
 
 interface Props {
   control: Control<AutomationRuleFormData>;
@@ -477,6 +487,9 @@ function ActionParamsRenderer({ control, index, actionName, formData, t }: Param
         />
       );
 
+    case 'update_custom_attribute':
+      return <CustomAttributeParam control={control} index={index} formData={formData} t={t} />;
+
     default:
       return null;
   }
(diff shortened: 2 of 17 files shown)


dpaes:
## Review — Changes requested 🔴 (EVO-1751, frontend half)

Reviewed together with crm #158. The wire contract matches the backend exactly, all 9 display types are covered, and i18n parity across the 6 locales is good. Two real issues:

**HIGH — the `actionRegistry.spec.ts` parity guard will fail:**
The registry now exposes 20 actions (includes `update_custom_attribute`), but the spec's hardcoded `BACKEND_ACTION_NAMES` is still 19 entries → `expect(ALL_ACTION_NAMES.sort()).toEqual(...)` **fails**. Because CI only runs Sourcery (no vitest), "tests pass" is self-reported and this spec ships red — nullifying the very front/backend parity guard it exists to provide. Add `update_custom_attribute` to `BACKEND_ACTION_NAMES` and confirm vitest green locally.

**MEDIUM — checkbox emits a string, diverging from the canonical boolean:**
`CustomAttributeValueInput` emits `'true'`/`'false'` (string) and the backend stores it verbatim in JSONB. But the canonical custom-attribute path stores a real boolean (`CustomAttributesForm` writes `!currentValue` and reads `Boolean(rawValue)`). Since `Boolean('false') === true`, a checkbox set to "unchecked" via automation will show **checked** in the contact/conversation sidebar (the EVO-1275 trap). Emit a real boolean for `display_type` checkbox (or align with the canonical producer).

LOW:
- `UPDATABLE_ATTRIBUTE_MODELS` is hardcoded and mirrors the backend switch with no CI gate linking them — manual sync, silent drift risk.

Leaving the PR open. The contract wiring itself is good work — just the guard spec and the checkbox type need fixing.

dpaes (approved):
## Re-review — APPROVED ✅

Evaluated as a pair: **crm #158 (AC2/backend)** + **frontend #172 (AC1/builder)**. Both base `develop`, CLEAN/MERGEABLE.

All round-1 findings are resolved and independently verified against the canonical code (direct read + a 3-agent adversarial panel, all `refuted: false`, high confidence):

- **H1 (HIGH)** — `update_custom_attribute` added to `BACKEND_ACTION_NAMES`; the front↔backend parity guard is green (vitest 8/8). ✅
- **M1 (MED)** — checkbox is now coerced to a **real boolean at the backend write boundary** (`cast_custom_attribute_value` by `attribute_display_type`). Verified correct: every canonical reader uses `Boolean(rawValue)` truthiness (`CustomAttributesForm.tsx:364/561`, `PipelineCustomAttributesForm.tsx:498`, chat header/sidebar), so a real boolean renders correctly for both `true` and `false` — and it actually fixes the latent `Boolean('false') === true` trap. Keeps the string wire contract; non-checkbox types stay verbatim (cast at read/query time, matching `filter_service` `::numeric`/`::date`). ✅
- **M2 (MED)** — both specs run green on an isolated test DB (15 examples, 0 failures). Confirmed the specs exercise the real dispatch (`send(action_name)` gated by `AutomationRule#actions_attributes`, no mocking). ✅
- **L2 (LOW)** — a contact-triggered `update_custom_attribute` targeting a conversation/pipeline attribute now records an explicit `warn` skip instead of a misleading `success` on a silent no-op. ✅

Deferred with sound justification: **L1** (front↔back `UPDATABLE_ATTRIBUTE_MODELS` drift — a real guard needs cross-repo CI tooling that doesn't exist; verified manually), **L3** (`pipeline_items.first` — consistent with the existing `create_pipeline_task`).

Additional verification: cross-repo wire contract matches char-for-char (`custom_attribute_{key,model,value}`, 3-model vocabulary, string value both sides); backend write is safe (definition-gated lookup, fixed JSONB columns, no mass-assignment); the automation-loop guard covers all three write paths (`Current.executed_by` set globally + `performed_by_automation?` / `contact.rb` guard / pipeline stage-change gate); AC1 populates end-to-end.

### Non-blocking notes (informational, not gating)
- Type heterogeneity: checkbox stored as a native boolean while number/currency/date stay as wire strings — internally consistent (query-time casts), noted for any future native-type reader.
- Pre-existing: a `pipeline_item_attribute` whose `attribute_key` is literally `services`/`currency` would merge over the reserved structured deal-value in `custom_fields` (also reachable via the existing sidebar; `attribute_must_not_conflict` doesn't block those keys). Not introduced here.

Squash-merging both and deleting the `feat/EVO-1751` branches. Submodule pointer bumps in the umbrella are handled in the release batch, not here.

PR #173, opened 2026-06-22
fix(frontend): insert canned response correctly (EVO-1685)

nickoliveira23:
## Summary
- Selecting a canned response (typing `/`, then click or Enter) was a silent no-op — the inserted text never reached the message input.
- **Root cause:** `handleSelectCannedResponse` sliced the editor content at `lastIndexOf('/')`, but `getContent()` returns **HTML** (`<p>/</p>`), so the slash matched the closing `</p>` tag — the result was mangled and the canned content dropped.
- **Fix:** extract `buildCannedResponseMessage()` that locates the `/` trigger in the **plain text** (not the HTML) and rebuilds the message.

> Note: the Linear card pointed at `RichTextEditor.setContent()` (updateState vs dispatch) as the cause. That is a real-but-secondary inconsistency, not the blocker — the handler already syncs the value manually, and fixing only `setContent` did not resolve the bug (verified). `RichTextEditor` is left untouched.

## Validation
- `frontend: pnpm test -- src/components/chat/message-input/buildCannedResponseMessage.spec.ts` (4 passing)
- `frontend: pnpm exec tsc -b --noEmit` (clean for changed files; 3 pre-existing `StageAutomationRules` errors are unrelated)
- `frontend: pnpm lint` on changed files (clean; pre-existing warnings only)
- Manual smoke: typing `/`, selecting a canned response by **click** and by **Enter** → text inserted in the input (confirmed).

## Changed Files
- `src/components/chat/message-input/buildCannedResponseMessage.ts` (new)
- `src/components/chat/message-input/buildCannedResponseMessage.spec.ts` (new)
- `src/components/chat/message-input/MessageInput.tsx`

## Linked Issue
- EVO-1685

## Summary by Sourcery

Fix canned response insertion in the chat message input by building the final message from plain text instead of the editor’s HTML content.

Bug Fixes:
- Ensure selecting a canned response correctly replaces the '/' trigger in the message input instead of silently failing when HTML is involved.

Enhancements:
- Extract message-building logic for canned responses into a reusable utility function for the message input.

Tests:
- Add unit tests covering the canned response message-building logic for different input scenarios.

### DIFF
--- a/src/components/chat/message-input/MessageInput.tsx
+++ b/src/components/chat/message-input/MessageInput.tsx
@@ -36,6 +36,7 @@ import AudioRecorder from '../audio';
 
 import { AIAssistanceButton } from '../ai-assistance';
 import { CannedResponsesList } from '../canned-responses';
+import { buildCannedResponseMessage } from './buildCannedResponseMessage';
 import { RichTextEditor, RichTextEditorRef } from '../rich-text-editor';
 
 import { ReplyMode, Message } from '@/types/chat/api';
@@ -168,13 +169,10 @@ const MessageInput: React.FC<MessageInputProps> = ({
 
   const handleSelectCannedResponse = useCallback(
     async (cannedResponse: CannedResponse) => {
-      const currentMessage = richEditorRef.current?.getContent() || '';
-
-      const slashIndex = currentMessage.lastIndexOf('/');
-      const newMessage =
-        slashIndex >= 0
-          ? currentMessage.substring(0, slashIndex) + cannedResponse.content
-          : cannedResponse.content;
+      const newMessage = buildCannedResponseMessage(
+        richEditorRef.current?.getContent() || '',
+        cannedResponse.content,
+      );
 
       richEditorRef.current?.setContent(newMessage);
       setCurrentEditorMessage(newMessage);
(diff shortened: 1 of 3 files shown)


dpaes (approved):
## Review — Approved ✅ (EVO-1685)

The fix is correct and the root-cause pivot supersedes the original card diagnosis. The real bug was `handleSelectCannedResponse` running `lastIndexOf('/')` over the **HTML** from `getContent()` (`<p>/</p>`) — the slash matched the closing `</p>`, mangling the content. The new `buildCannedResponseMessage()` locates the `/` trigger in **plain text**. `RichTextEditor` is left untouched, which is correct: the `setContent` `updateState` vs `dispatch` concern from the card is **not** a bug — `updateState` is the right ProseMirror API and repaints the editor (same path as `clear()`).

Minor (non-blocking):
- The `Sourcery review` FAILURE (CI UNSTABLE) is a **verified false positive**: `wrapper.innerHTML` runs on a **detached** div, the content is the editor's own serialized doc over a restricted schema (no img/script), and only `.textContent` is read. Merging despite the red advisory check for that reason. Consider switching to `DOMParser`/tag-strip to keep the gate green next time.
- Plain-text flatten drops formatting before the `/` (low impact); the 4 vitest specs don't run in CI (Sourcery-only) — self-reported.

Approving and squash-merging. Nice diagnostic work on the root cause.

PR #174, opened 2026-06-22
feat(frontend): polish product create/edit form + catalog stock column (EVO-1783)

nickoliveira23:
## Summary
Phase 1 UX polish of the product create/edit form (`ProductModal`) + the catalog table. **Frontend-only** — the community `Product` model/API/validations already cover everything (no backend change).

- **Stock only for physical** products (general field + per-variant); stock is nulled for digital on submit.
- **Visible validation** — submit is never a silent disabled button: clicking an invalid form reveals field errors and jumps to the General tab; required-field legend; empty price stays `null` (not `0`).
- **Server 422 field errors** (e.g. SKU uniqueness) surfaced **inline** in the form instead of a generic toast.
- **Price + currency on one row**; field order per the UX spec.
- **i18n** — hardcoded placeholders + currency labels across all 6 locales.
- **a11y** — `htmlFor`/`id` on every select + variant input; `aria-required`/`aria-invalid`.
- **Media tab removed** — image upload is not wired end-to-end (backend `attach_images` drops raw multipart files, expects ActiveStorage signed_ids). A code comment documents how to restore it; a real upload flow is a separate follow-up.
- **Catalog table: Stock column** — digital → "—"; untracked → "—"; physical → number; with variants → summed; `0` → "Out of stock" (red).

## Security
No new endpoints, no authZ change. Client validation mirrors server rules for fast feedback; the backend stays the source of truth (its 422 is surfaced inline).

## Test plan
- `frontend: pnpm exec vitest run src/components/products/` — 11 tests pass
- `frontend: pnpm exec eslint <changed files>` — 0 warnings
- `frontend: pnpm exec tsc -b --noEmit` — no errors in changed files
- Manual smoke: create physical/digital product (stock conditional), trigger required validation + a duplicate-SKU 422 (inline), verify the catalog Stock column.

## Changed Files
- `src/components/products/ProductModal.tsx`
- `src/components/products/ProductModal.spec.tsx`
- `src/components/products/ProductsTable.tsx`
- `src/components/products/ProductsTable.spec.tsx`
- `src/components/products/productStock.ts`
- `src/pages/Customer/Settings/Products/Products.tsx`
- `src/i18n/locales/{en,pt-BR,es,fr,it,pt}/products.json`

## Out of scope / follow-ups
- Phase 2 (labels → chips, variant grid header, loading state, post-create visibility) + Phase 3 (single-screen redesign, deep-link `/products/:id`).
- Real product image upload (needs backend signed_ids / direct-upload flow).

## Linked Issue
- EVO-1783

### DIFF
--- a/src/components/products/ProductModal.spec.tsx
+++ b/src/components/products/ProductModal.spec.tsx
@@ -1,0 +1,65 @@
+import { describe, it, expect, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import ProductModal from './ProductModal';
+import type { Product } from '@/types/products';
+
+class ResizeObserverPolyfill {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
+globalThis.ResizeObserver = globalThis.ResizeObserver ?? (ResizeObserverPolyfill as never);
+const proto = HTMLElement.prototype as unknown as Record<string, unknown>;
+if (!proto.hasPointerCapture) proto.hasPointerCapture = () => false;
+if (!proto.scrollIntoView) proto.scrollIntoView = () => {};
+
+const baseProduct = (overrides: Partial<Product>): Product => ({
+  id: 'p1',
+  name: 'Prod',
+  kind: 'physical',
+  default_price: 10,
+  currency: 'BRL',
+  status: 'active',
+  variants: [],
+  images: [],
+  ...overrides,
+});
+
+const noop = () => {};
+const onSubmit = vi.fn(async () => {});
+
+describe('ProductModal (EVO-1783 Phase 1)', () => {
+  it('hides the stock field for digital products (AC4)', () => {
+    render(
+      <ProductModal open product={baseProduct({ kind: 'digital' })} loading={false} onOpenChange={noop} onSubmit={onSubmit} />,
+    );
+    expect(document.getElementById('p-stock')).toBeNull();
+  });
+
+  it('shows the stock field for physical products (AC4)', () => {
+    render(
+      <ProductModal open product={baseProduct({ kind: 'physical' })} loading={false} onOpenChange={noop} onSubmit={onSubmit} />,
+    );
+    expect(document.getElementById('p-stock')).not.toBeNull();
+  });
+
+  it('disables submit when required fields are empty on create', () => {
+    render(<ProductModal open product={null} loading={false} onOpenChange={noop} onSubmit={onSubmit} />);
+    const button = screen.getByText('actions.create').closest('button') as HTMLButtonElement;
+    expect(button.disabled).toBe(true);
+  });
+
+  it('renders a server field error inline (AC4 — SKU uniqueness)', () => {
+    render(
+      <ProductModal
+        open
+        product={baseProduct({})}
+        loading={false}
+        errors={{ sku: 'has already been taken' }}
+        onOpenChange={noop}
+        onSubmit={onSubmit}
+      />,
+    );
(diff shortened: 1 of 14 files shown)


dpaes:
## Review — Changes requested 🔴 (EVO-1783)

Solid, well-scoped UX work (visible validation, price+currency on one row, correct digital/physical stock semantics — verified no leak in either direction, 6-locale i18n parity, a11y). One HIGH issue leaves part of the AC unmet end-to-end, so I'm sending this back.

**HIGH — inline 422 mapping doesn't match the backend contract (AC4 not met):**
`toFieldErrors` (`Products.tsx:116`) expects `details` as `{field, message}[]` or `{field: [msgs]}`, but the backend (`products_controller#validation_error_response:213`) sends `record.errors.full_messages` — a **flat array of strings**. In the `Array.isArray` branch the `'field' in d` guard rejects strings → `fieldErrors` ends up `{}` → it falls back to the generic `toast.error`. So the AC4 anchor example (inline SKU uniqueness) **never renders under the field**. The PR specs feed a pre-formatted `errors` prop and never exercise `toFieldErrors` against the real payload, so the gap is invisible in tests.

Fix: align the seam — either the backend emits per-field `errors.messages` (or `[{field, message}]`), or the frontend maps `full_messages`→field. Add a `toFieldErrors` spec with the real payload to lock it.

LOW:
- Orphan Media i18n keys after the tab removal (intentional — fine to keep as a re-enable footprint).
- Pasted price can become `NaN` (narrow edge) — guard `Number.isNaN`.

Leaving the PR open. Happy to re-review once the 422 seam is addressed.

dpaes (approved):
## Re-review — APPROVED ✅ (round 2)

The round-1 HIGH (AC4 — the inline 422 field error never reaching the form) is fully resolved end-to-end, verified across both repos (paired with crm#160).

**Root cause (corrected).** The round-1 diagnosis (a 422 with a flat array the parser couldn't key) was wrong about the *mechanism*. The products endpoint never returned 422 — every error path called `error_response` with keyword args (`code:`, `message:`) against the positional signature `(code, message, details:, status:)`, raising `ArgumentError` → **500**. Identical user-facing symptom (generic toast), which is why static review couldn't tell them apart. Well diagnosed and closed on both sides.

**AC4 chain verified:**
- crm `format_validation_errors` emits `[{field, messages, full_messages}]` (snake_case `field`) under `error.details` (422).
- `extractError` reads `response.data.error.details`; `toFieldErrors` maps `field → message ?? messages[0] ?? full_messages[0]`.
- Field keys (`sku`, `name`, `default_price`, `purchase_url`) match the form's `fieldError(...)` keys → inline render.
- Backed by a real model rule: `Product validates :sku, uniqueness: true, allow_blank: true`.
- `productErrors.spec.ts` now exercises the **real** structured payload — the exact round-1 gap (specs previously used pre-formatted errors).

**Round-1 lows:** L1 (NaN from pasted price/stock) fixed via `toNumberOrNull` across product + variant inputs. L2 (orphan Media i18n keys) deferred/intentional.

**Non-blocking notes:** dead i18n keys `validation.skuTaken` and `validation.variantPriceCurrency` were added to all 6 locales but are unused — trivial cleanup. i18n parity across all 6 locales and the a11y additions are confirmed.

**CI caveat:** frontend CI does not run vitest/tsc/eslint — "27 passed / tsc clean" is self-reported; the new specs would catch the regression if run.

Approving and merging (squash). The umbrella submodule-pointer bump is intentionally left for the batched release chore.

PR #177, opened 2026-06-22
fix(frontend): persist audio transcription toggle (EVO-1681)

nickoliveira23:
## Summary
Fix the Account Settings **"Audio Transcription"** toggle not persisting after reload. The handler sent `audio_transcriptions` at the **root** of the `PATCH /account` payload, but the backend `account_controller` permits it only under `settings: {}` — so it was silently dropped and reverted on F5. Now nested under `settings` (preserving existing keys) + account/appDataStore cache refresh, mirroring `handleMaskContactPiiToggle` (EVO-1551).

Frontend-only, 1 file, zero backend change.

## Test plan
- **Manual smoke (confirmed):** enable "Audio Transcription" → F5 → stays ON (previously reverted to OFF). The sibling `Mask Contact Data` toggle (same `settings` merge) keeps working.
- `frontend: pnpm exec tsc -b --noEmit` (clean for the changed file)
- `frontend: pnpm exec eslint` (no new errors; the file has 1 pre-existing `any` + 1 pre-existing exhaustive-deps warning, both unrelated to this change)

> Note for reviewers: the toggle is only visible when an Evolution integration is configured (`hasEvolutionConfig || hasEvolutionGoConfig` from `/global_config`). To surface it, set `EVOLUTION_GO_API_URL` + `EVOLUTION_GO_ADMIN_SECRET` (any non-empty values) in Admin → Channel Config.

## Changed Files
- `src/pages/Customer/Settings/Account/AccountSettings.tsx`

## Linked Issue
- EVO-1681

## Summary by Sourcery

Bug Fixes:
- Fix Audio Transcription toggle not persisting after page reload by sending it under account settings and refreshing account state caches.

### DIFF
--- a/src/pages/Customer/Settings/Account/AccountSettings.tsx
+++ b/src/pages/Customer/Settings/Account/AccountSettings.tsx
@@ -265,10 +265,19 @@ export default function AccountSettings() {
 
   const handleAudioTranscriptionToggle = async (enabled: boolean) => {
     try {
-      await accountService.updateAccount({
-        audio_transcriptions: enabled,
+      const updated = await accountService.updateAccount({
+        settings: {
+          ...(account?.settings ?? {}),
+          audio_transcriptions: enabled,
+        },
       });
+      setAccount(updated);
       setFormData(prev => ({ ...prev, audioTranscriptions: enabled }));
+      try {
+        await useAppDataStore.getState().fetchAccount(true);
+      } catch (cacheError) {
+        console.warn('Failed to refresh appDataStore.account cache:', cacheError);
+      }
       toast.success(
         enabled
           ? t('messages.success.audioTranscriptionEnabled')


dpaes (approved):
## Code review — EVO-1681 ✅ Approved

Minimal, correct fix. The handler now sends `audio_transcriptions` nested under `settings` (preserving existing keys) + `setAccount(updated)` + non-blocking `fetchAccount(true)` cache refresh — a byte-for-byte mirror of `handleMaskContactPiiToggle`, which is already live in `develop` (EVO-1551) and confirmed working.

**Verified against the code (not just self-report):**
- Bug confirmed real: `origin/develop` sends `audio_transcriptions` at the payload root → dropped by the controller permit-list (`name/domain/support_email/locale` + `settings:{}`).
- `updateAccount` returns `Promise<Account>` → `setAccount(updated)` is type/runtime-correct.
- Types line up: `Account.settings.audio_transcriptions?` and `UpdateAccount.settings.audio_transcriptions?` both exist.
- `appDataStore.fetchAccount(forceRefresh?: boolean)` exists → `fetchAccount(true)` is valid.
- JSX wiring intact (`Switch checked={formData.audioTranscriptions}` / `onCheckedChange={handleAudioTranscriptionToggle}`), toggle gated by `isOnEvolutionCloud`.
- i18n keys `audioTranscription{Enabled,Disabled,Failed}` present in all 6 locales; no new keys.
- AC ("enable → F5 → stays ON") met.

**Findings — 0 Blocker / 0 High / 0 Medium / 4 Low (none blocking):**
- L1 — No automated test (page hard to render-test; sibling `mask_pii` also untested; CI doesn't run vitest). Acceptable.
- L2 — `settings` merge reads local `account?.settings`; cross-tab/device stale window could revert other keys. Inherited from the `mask_pii` pattern.
- L3 — DRY: the audio & mask handlers are now near-identical; could extract a shared `updateAccountSetting(key, value)`.
- L4 — Footgun in the `UpdateAccount` type: the root-level `audio_transcriptions?` / `mask_contact_pii?` fields remain — these are exactly the dropped-at-root fields that caused this bug, so a future dev could reintroduce it. (Keep the root-level `auto_resolve_*` — `handleAutoResolveSubmit` sends them at root.)

CI note: frontend CI is Sourcery-only (no tsc/vitest gate), so the "tsc clean" claim is self-reported; static type cross-check independently confirms it compiles.

Approving and squash-merging into `develop`. Submodule pointer bump in the umbrella is intentionally left for the batched release chore.

PR #178, opened 2026-06-23
feat(journey): structured webhook body builder (EVO-1742)

daniloleonecarneiro:
## Summary

Replaces the raw hand-typed JSON body in the Journey **Send Webhook** node with an n8n-style **structured key/value builder** (typed `string` / `number` / `boolean`), mirroring the existing `WebhookHeadersConfig` pattern already in the node. A raw **"advanced" mode** toggle is kept for nested/complex bodies, and the body is now **validated in the editor** instead of failing silently at journey runtime (`SendWebhookPanel.tsx` previously only `JSON.parse`d in the Test flow).

**Frontend-only.** The new `bodyStructured` rows are serialized into the existing `body` string, which stays the wire source-of-truth — so the live executor and the Test tab keep working with **zero backend change**.

### What changed
- **Node data:** new `bodyStructured: WebhookBodyField[]` (`{id,key,value,type}`) + `bodyMode: 'structured' | 'raw'` (`SendWebhookNode.tsx`).
- **Pure helpers** (`webhookBody.ts`): `serializeBody`, `tryParseToFields` (with a nested-JSON guard), `validateFields`, `coerceJsonValue`, `isValidJsonBody`, `getEffectiveBodyMode`.
- **Builder** (`WebhookBodyBuilder.tsx`): typed key/value rows; type column for `json`, hidden for `form`.
- **Body tab** (`WebhookBodyConfig.tsx`): hosts the Structured⇄Raw toggle + round-trip; JSON templates only in raw mode.
- **Validation** wired into the existing `dirty && isValid` save gate (`SendWebhookPanel.tsx`): invalid JSON / blank key / duplicate key block save.
- **i18n:** new keys across all 7 locales + `evo1742Keys` registered in `journey-parity.spec.ts`.

### Design notes / back-compat
- **Existing raw-body journeys** open in **raw mode untouched** — the effective mode is derived **render-only** (no mount-time `onChange`, so no spurious dirty / no-op re-serialization of legacy bodies).
- A persisted `bodyMode` always wins on reopen; new structured-capable nodes default to structured.
- **Nested-JSON guard:** `raw → structured` is refused (with an inline hint) when the body has nested objects/arrays, so we never lossily flatten.
- A `number`-typed field holding a `{{variable}}` is preserved as a string token (the executor substitutes it later) rather than emitting `NaN`. JSON validity is checked variable-tolerantly.
- Builder covers `json` + `form`; `text`/`xml` keep the raw textarea.

## Test plan
- `frontend: npx tsc --noEmit` — exit 0
- `frontend: npx eslint <changed files>` — clean
- `frontend: npx vitest run` — **55 green**: `webhookBody.spec.ts` (26 pure-fn), `WebhookBodyConfig.spec.tsx` (5 component: build field-by-field, raw back-compat + no mount onChange, mode round-trip, nested guard, form urlencoded), `journey-parity.spec.ts` (24).
- Full `src/components/journey` suite: 230 passing. The single failing file (`tokens.contrast.spec.ts`) is a **pre-existing** `colorjs.io` transform error — confirmed it fails identically on clean `develop`, unrelated to this change.

> **Follow-up owed (separate card):** teach the journey webhook **executor to read `bodyStructured`** (typed values) directly. Until then `body` remains the wire format, which this PR keeps serialized.

### DIFF
--- a/src/components/journey/nodes/actions/send-webhook/SendWebhookNode.tsx
+++ b/src/components/journey/nodes/actions/send-webhook/SendWebhookNode.tsx
@@ -14,6 +14,15 @@ export interface WebhookResponseMapping {
   description?: string;
 }
 
+export type WebhookBodyValueType = 'string' | 'number' | 'boolean';
+
+export interface WebhookBodyField {
+  id: string;
+  key: string;
+  value: string; // raw text; preserves {{variables}} verbatim
+  type: WebhookBodyValueType;
+}
+
 export interface SendWebhookNodeData {
   label: string;
   description?: string;
@@ -22,6 +31,8 @@ export interface SendWebhookNodeData {
   headers?: WebhookHeader[];
   body?: string;
   bodyType?: 'json' | 'form' | 'text' | 'xml';
+  bodyStructured?: WebhookBodyField[];
+  bodyMode?: 'structured' | 'raw';
   timeout?: number;
   retryAttempts?: number;
   authenticationType?: 'none' | 'bearer' | 'basic' | 'api_key';
--- a/src/components/journey/nodes/actions/send-webhook/SendWebhookPanel.tsx
+++ b/src/components/journey/nodes/actions/send-webhook/SendWebhookPanel.tsx
@@ -7,6 +7,11 @@ import { FlowFeedbackBanner } from '@/components/journey/_ui';
 import { VariableSelect } from '@/components/journey/environment-manager';
 import { useLanguage } from '@/hooks/useLanguage';
 import {
+  getEffectiveBodyMode,
+  isValidJsonBody,
+  validateFields,
+} from './components/webhookBody';
+import {
   WebhookBasicConfig,
   WebhookHeadersConfig,
   WebhookBodyConfig,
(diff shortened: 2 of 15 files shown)


dpaes (approved):
## ✅ Approved — EVO-1742

Reviewed the structured webhook-body builder against the three acceptance criteria and verified the build claims locally (frontend CI only runs Sourcery, so `tsc`/`vitest` are not gated there).

**Acceptance criteria — all met**
- **AC1 (build field-by-field):** `WebhookBodyBuilder` renders typed key/value/type rows backed by `VariableInput`, serializing into `body` on every edit. `json` (typed) + `form` (no type column); `text`/`xml` keep the raw textarea, which is correct for free-form payloads.
- **AC2 (caught in editor, not at runtime):** validation flows into `getValidationStatus()` → `isValid`, and `NodeConfigModal` disables Save via `disabled={!dirty || loading || saveDisabled}` with `dirty={dirty && isValid}` — I confirmed this genuinely blocks Save. Structured rows validate `blankKey`/`duplicateKey`; raw JSON uses the variable-tolerant `isValidJsonBody`.
- **AC3 (back-compat):** `getEffectiveBodyMode` is render-only (no mount-time `onChange` → no spurious dirty), legacy raw bodies open untouched in raw mode, and the nested-JSON guard refuses lossy `raw→structured` flattening.

**Verified locally**
- `vitest` **55/55 green** (26 pure-fn + 24 parity + 5 component).
- `tsc -b` clean on the **merged** result (develop + this branch) → exit 0. The 3 `StageAutomationRules.tsx` errors seen when the branch is checked out in isolation are pre-existing, untouched by this PR, and already fixed on `develop`; they disappear on merge.
- i18n: 6 journey locales updated and covered by `journey-parity.spec.ts`, new keys translated (no English leakage).

**Non-blocking notes (LOW/INFO) — fine to land as-is**
1. *(LOW)* Emptying the last JSON structured row serializes to `{}` (non-empty), so the panel auto-flips to raw mode; `form` doesn't (serializes to `''`). Recoverable, minor UX asymmetry.
2. *(LOW)* `number`/`boolean` rows whose value is a `{{variable}}` lose their declared type on a structured⇄raw round-trip (re-inferred as `string`); wire output is identical, so cosmetic only.
3. *(LOW)* `form` values containing `&`/`=` round-trip incorrectly (no percent-encoding — a deliberate choice to preserve `{{variables}}`, matching prior raw behavior).
4. *(INFO)* `duplicateKey` is also enforced for `form`, which urlencoded technically allows.
5. *(INFO)* A `number`-typed `{{variable}}` always serializes quoted on the wire — the documented decision; the real fix is the follow-up that teaches the executor to read `bodyStructured`.

Solid, well-tested, careful back-compat. Approving and merging.

PR #180, opened 2026-06-23
feat(journey): pre-activation validation framework (EVO-1744)

daniloleonecarneiro:
## Summary

Generalizes the lone exit-node check (EVO-1692) into a **per-node pre-activation validation framework** so an incoherent journey can't be shipped. Editor-side mirror of the runtime trigger↔action contract (EVO-1741); the terminal-path check is **folded in**, not duplicated.

**Engine** (`src/utils/journeyFlowValidation.ts` + `journeyValidators/`): `validateJourney(nodes, edges) → { errors, warnings, byNodeId, isActivatable }`. Pure (no React/fetch) → runs in the editor **and** off persisted `flowData` from the journey list/modal.

- **Required-config (errors):** headless field-presence validators per node type. Rich/async checks (balanced expressions, required template vars) stay in the panels, which already gate their own save — the framework is the "never configured" backstop.
- **Trigger↔action coherence (warnings):** a conversation-requiring action reachable from a trigger that provides no conversation. Per-trigger semantics (one session per trigger). Reads `data.triggerType` + the event catalog category.
- **Terminal-path/dangling (warnings):** folded in from EVO-1692.

**Hybrid enforcement:** errors **block** activation; coherence + dangling are **warnings** (allow, but surfaced).

**Surfaces:**
- Hard block in the list toggle (`handleToggleJourney`) and modal save, both on persisted `flowData`; branch on `isActive` before the bodyless toggle; deactivation always allowed.
- Editor: pre-activation summary banner + per-node markers via a validation **context** feeding `NodeValidationBadge` in the shared `BaseFlowNode` — never written into node `data`, so no autosave/persist side effects.

## Test plan

- `npx vitest run src/utils/journeyFlowValidation.spec.ts src/utils/journeyValidators` → **25/25** (engine, validators, coherence; covers AC1–AC5/AC7).
- `npx tsc -b` clean.
- AC1 missing config blocks list activation · AC2 contact-trigger→conversation-action warns (non-blocking) · AC3 per-node markers + summary banner · AC4 dangling folded into the engine · AC5 hybrid (only errors gate) · AC7 happy path activates · AC8 deactivate never validated.

## Notas

- The conversation-requiring set (13) mirrors evo-flow `base.node.ts requiredContext: 'conversation'` (FE `-node` suffix; runtime `snooze-conversation` = FE `defer-conversation-node`).
- Validators are intentionally narrower than panel `isValid` (presence only) — under-validate, never over-block.
- **Out of scope:** cycle-without-exit detection (inherited blind spot from EVO-1692) → tracked in EVO-1857.

Closes EVO-1744.

### DIFF
--- a/src/components/base/BaseFlowNode.tsx
+++ b/src/components/base/BaseFlowNode.tsx
@@ -1,9 +1,10 @@
-import { Handle, Position, useEdges } from "@xyflow/react";
+import { Handle, Position, useEdges, useNodeId } from "@xyflow/react";
 import React from "react";
 import { useLanguage } from '@/hooks/useLanguage';
 import { cn } from "@/lib/utils";
 import { useDnD } from "@/contexts/DnDContext";
 import { ArrowRight } from "lucide-react";
+import { NodeValidationBadge } from '@/components/journey/_ui/NodeValidationBadge';
 
 // Tipos para configuração do node - COMPATÍVEL COM BaseNode ATUAL
 export interface BaseFlowNodeProps {
@@ -178,6 +179,10 @@ export function BaseFlowNode({
   const { t } = useLanguage('common');
   const { pointerEvents } = useDnD();
   const edges = useEdges();
+  // EVO-1744: resolve the node id (explicit prop or the React Flow context) so
+  // the validation badge can look up its issues without each node forwarding it.
+  const hookNodeId = useNodeId();
+  const resolvedNodeId = nodeId ?? hookNodeId;
 
   // Verificar se o source handle está conectado - compatível com BaseNode atual
   const isHandleConnected = (handleId: string) => {
@@ -216,6 +221,9 @@ export function BaseFlowNode({
       }}
       data-is-executing={isExecuting ? "true" : "false"}
     >
+      {/* EVO-1744: pre-activation validation marker (error/warning) */}
+      <NodeValidationBadge nodeId={resolvedNodeId} />
+
       {/* Target Handle - mantém comportamento exato do BaseNode atual */}
       {hasTarget && (
         <Handle
--- a/src/components/journey/JourneyModal.tsx
+++ b/src/components/journey/JourneyModal.tsx
@@ -13,6 +13,8 @@ import {
 } from '@evoapi/design-system';
 import { journeyService } from '@/services';
 import { Journey, TriggerType } from '@/types/automation';
+import type { Node, Edge } from '@xyflow/react';
+import { validateJourney } from '@/utils/journeyFlowValidation';
 import { toast } from 'sonner';
 import { useLanguage } from '@/hooks/useLanguage';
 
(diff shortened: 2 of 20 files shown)


dpaes (approved):
## Code review — EVO-1744 · Approved ✅

Reviewed against the actual PR head (`873ce84`) in an isolated worktree (my local submodule was on an unrelated branch, which would have produced false findings against a stale modal/panels). Engineering is clean and all acceptance criteria are met.

### Acceptance criteria

| AC | Status | Evidence |
|----|--------|----------|
| **AC1** — activation blocked when a node is missing required config | ✅ | `errors → isActivatable=false`; hard block in the list toggle (`handleToggleJourney`) and the modal save (off→on only), both over persisted `flowData` |
| **AC2** — incompatible trigger↔action pairs warned | ✅ | `validateTriggerActionContext` emits a non-blocking **warning**, per-trigger semantics |
| **AC3** — per-node markers + pre-activation summary | ✅ | `NodeValidationBadge` fed by context (never written into node `data` → no autosave/persist side effects) + summary banners |
| **AC4** — EVO-1692 exit check folded in, not duplicated | ✅ | `terminalPathIssues` reuses `validateJourneyTerminalPaths`; no leftover call sites in the editor |

### Independently verified (the failure modes this codebase tends to hide)

- **Validator field names vs what panels actually persist** — the highest risk (a field read flat that is actually nested → false-positive error blocking a correctly-configured journey). All 24 required-config validators match **flat** on `data`; the `formDataOptions` nodes (assign-*, pipeline, canned, transcript, priority, email-team) write their ids at the top level of `data`, not inside the wrapper. **No mismatch.**
- **All 27 node `type` strings** match the `nodeTypes` registry in `JourneyFlowEditor` (no silent no-op from a misspelled type).
- **events-manifest + triggerType** — `getEvent(name)?.category` returns `'conversation'`/`'message'` as assumed; all 9 lowercase-camelCase `triggerType` values exist and are handled.
- **`flowData` is populated per item in the journey list** (`JourneyFlowData` requires `nodes`/`edges`; `JourneyResponseDto` includes `flowData`) — so AC1 enforcement from the list is real, not a no-op.
- **Compiles** (FE CI does not run tsc/vitest): `labelFor` is defined, imports are used, the `JourneyValidationProvider` JSX is balanced, and the new i18n keys exist across all 6 locales.

### Non-blocking follow-ups (none gate this merge)

1. **Create-from-modal activates without validation** — a brand-new journey created with `isActive: true` (trigger-only default flow) bypasses the gate (`journey?.id` is falsy). Harmless (empty journey does nothing; the flow is built in the editor afterward), but worth noting.
2. **AC1-from-list relies on a cross-repo contract with no guard** — if evo-flow ever trims `flowData` from the list response for performance, list-side blocking silently degrades to "always activatable". Currently holds; a defensive note is enough.
3. **`CONVERSATION_REQUIRING_NODES` (13) is a hand-maintained mirror** of evo-flow `requiredContext: 'conversation'` (EVO-1741) with no shared source/CI guard → can drift from the runtime (same pattern as the contact_events manifest).
4. **Required-config validates unreachable/orphan nodes too** — a disconnected misconfigured node blocks activation. Likely desirable, but it is a behavioral choice worth calling out.
5. **Orphaned i18n keys** — `danglingExitBanner` / `danglingExitWarning` no longer have code consumers in any of the 6 locales (replaced by `summary*` / `saveIssues`). Trivial cleanup.
6. **Minor token inconsistency** — the badge uses raw `text-amber-500` for warnings vs the `text-flow-feedback-error-fg` token for errors. Legible, just inconsistent.

Note: the "25/25 vitest + `tsc -b` clean" claims are self-reported; contracts were verified statically here, the suite was not executed (FE CI does not run it either).

Cycle-without-exit detection is correctly carved out to EVO-1857.

PR #161, opened 2026-06-15
fix(layout): re-expose Journeys, Campaigns and Segments in the community sidebar (EVO-1687)

nickoliveira23:
## Summary
PR #84 commented Journeys, Campaigns and Segments out of the community sidebar. Per Davidson's decision (EVO-1568) they are core open-source/Community features, so this re-exposes them in `menuItems.ts`:
- Uncomment the `Filter` / `Megaphone` / `Route` icon imports.
- Uncomment the **Journeys** (`/journeys`), **Campaigns** (`/campaigns`) and **Segments** (`/settings/segments`) menu entries.
- **Tutoriais** stays out (per Daniel's note on the card — removed on purpose).

## Why this is safe / complete
- `menuItems.ts` is the single sidebar source (Sidebar/MenuItem consume it) — no other menu config to touch.
- Routes already exist (`routes/index.tsx`: `/journeys`, `/campaigns`, `/settings/segments`).
- i18n keys present in all 6 locales (`layout.json`).
- Items are permission-gated by `resource` (journeys/campaigns/segments) — unchanged from before #84.
- No edition/community flag re-hides them.

## Validation
- `tsc -b --noEmit` → no errors in the changed file
- `eslint src/components/layout/config/menuItems.ts` → clean

## Changed Files
- `src/components/layout/config/menuItems.ts`

## Linked Issue
- EVO-1687

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Re-expose core customer journey, campaign, and segment items in the community sidebar navigation.

New Features:
- Restore Journeys and Campaigns entries to the customer section of the sidebar when the user has appropriate permissions.
- Restore the Segments settings entry under customer settings in the sidebar when the user has appropriate permissions.

PR #181, opened 2026-06-23
fix(journey): persist Update Custom Attribute node value by attribute_key (EVO-1850)

daniloleonecarneiro:
## Summary
The journey **Update Custom Attribute** node persisted the contact attribute keyed by `attribute_display_name`, while everything else (CRM contact UI/API/import, and the Conditional node from EVO-1837) keys `custom_attributes` by `attribute_key` (the slug). So "set an attribute, then branch on it" silently failed — the Conditional read `{{contact.customAttributes.<attribute_key>}}` → `undefined`.

This PR makes the node persist `attributeName` as the **`attribute_key` slug** (the value the evo-flow runtime already uses as the CRM api key), and adds a dedicated `attributeDisplayName` field carrying the human-readable name for the UI (panel preview + canvas card), with a fallback to `attributeName` for legacy nodes.

## Test plan
- New `UpdateCustomAttributePanel.spec.tsx` (vitest) — asserts the saved node-data has `attributeName = slug` and `attributeDisplayName = display name`, and that the UI shows the display name (not the slug). **2/2 green.**
- `tsc -b` clean.

## Notas
- Part of EVO-1850 (3 repos): this FE PR + evo-flow (runtime merge/slug) + evo-ai-crm-community (backfill migration).
- Write-side counterpart to EVO-1837 (read side).

## Summary by Sourcery

Ensure the journey Update Custom Attribute node persists custom attributes by slug while displaying human-readable labels in the UI.

Bug Fixes:
- Fix incorrect persistence of custom attribute updates that previously used the display name instead of the attribute_key slug, enabling downstream conditionals to read the updated value correctly.

Enhancements:
- Introduce a separate attributeDisplayName field so node previews and panels show the human-readable attribute label while keeping attributeName as the slug for API integration.

Tests:
- Add unit tests for UpdateCustomAttributePanel to verify node data stores attributeName as attribute_key, uses attributeDisplayName for UI, and hides the raw slug in the preview.

### DIFF
--- a/src/components/journey/nodes/actions/update-custom-attribute/UpdateCustomAttributeNode.tsx
+++ b/src/components/journey/nodes/actions/update-custom-attribute/UpdateCustomAttributeNode.tsx
@@ -6,7 +6,10 @@ export interface UpdateCustomAttributeNodeData {
   label: string;
   description?: string;
   attributeId?: string;
+  // attributeName carries the attribute_key (slug) used as the CRM api key.
   attributeName?: string;
+  // attributeDisplayName is the human-readable label, for UI only (EVO-1850).
+  attributeDisplayName?: string;
   attributeDisplayType?: string;
   newValue?: string;
 }
@@ -44,7 +47,7 @@ export function UpdateCustomAttributeNode({ selected, data, id }: UpdateCustomAt
     }
 
     return t('panels.updateCustomAttribute.updates', {
-      attribute: data.attributeName,
+      attribute: data.attributeDisplayName || data.attributeName,
       value: data.newValue,
     });
   };
@@ -91,7 +94,7 @@ export function UpdateCustomAttributeNode({ selected, data, id }: UpdateCustomAt
             <div className="flex items-center gap-2 mt-1">
               <span className="text-xs">{getTypeIcon()}</span>
               <span className="text-xs text-pink-700 dark:text-pink-300 font-medium">
-                {data.attributeName}
+                {data.attributeDisplayName || data.attributeName}
               </span>
               <span className="text-xs text-pink-600 dark:text-pink-400">→ {data.newValue}</span>
             </div>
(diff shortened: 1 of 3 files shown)


dpaes:
EVO-1850 review: this FE change is sound — persists `attribute_key` as `attributeName` + new `attributeDisplayName` for UI, with graceful fallback for legacy nodes, and the spec proves slug ≠ display divergence. ✅

Not merging this round only because the card is held by a **data-loss blocker in the evo-flow side** (evo-flow-community #81): the read-modify-write merge reads `contact.customAttributes` off the raw snake_case `ContactDto` (real field is `custom_attributes`), so it never reads existing attributes and still wipes sibling `custom_attributes` on each run. Details on the card (EVO-1850) and on #81. This PR can ride along once #81 is fixed.

One non-blocking note for later: legacy journey nodes saved before this fix keep `attributeName = <display name>` unless the user re-selects the attribute (a plain re-save doesn't rewrite it), so they'd keep writing display-name keys. Worth considering resolving the slug from `attributeId` at runtime in evo-flow. Tracked in the card.

dpaes (approved):
## ✅ Approved (round 2)

Producer side verified. The panel now persists `attributeName = attribute_key` (the slug) and carries the human label in the new `attributeDisplayName` field; preview/canvas fall back to `attributeName` for legacy nodes. The spec deliberately uses a display name ≠ slug and proves the node-data stores the slug while the UI shows the display name. **AC1 / AC3 met.**

One note for the record (non-blocking, N1): there is **no load-time re-derivation** of `attributeName` from `attributeId` — the load effect only copies `data`, and `selectedAttribute` feeds the preview only. So a journey saved before this fix (with a display-name `attributeName`) keeps the stale key unless the user re-selects the attribute. The robust fix is runtime slug resolution from `attributeId` in evo-flow; the one-shot CRM backfill (#161) heals existing data. Tracked as accepted-gap / follow-up.

PR #182, opened 2026-06-23
fix(frontend): functional pagination, search & sort on config screens (EVO-1860)

nickoliveira23:
## Summary
Make the list controls on the three Settings config screens (Labels, Canned Responses, Custom Attributes) actually work. They were cosmetic: Labels and Canned Responses silently fetched only the first 20 rows (no `per_page` sent) and the pagination / sort / per-page controls never acted on the data.

- Load the full catalog in one request (`SETTINGS_LIST_FETCH_SIZE = 500` for the `apply_pagination`-capped Labels/Canned endpoints) and do search / sort / pagination client-side.
- Page slice is applied before the table; column sort actually reorders rows; the items-per-page selector is wired (the pagination wrappers were dropping `onPerPageChange`).
- Remove the destructive Labels search that overwrote `state.labels` (lost items when narrowing then widening the query).
- Distinct "no results" empty state on a non-matching search (shared `common:base.noResults`, all 6 locales) instead of the misleading "create first" CTA.
- Standardize table-to-pagination spacing across the three screens.

## Approach note
Rescoped to **client-side** (frontend-only, no backend changes) — these are bounded config catalogs (labels / canned / attribute definitions, realistically dozens), unlike Contacts/Conversations which justify the server-side pattern. Soft cap = 500; if a catalog ever exceeds it, migrate that screen to the Contacts server-side pattern. Full decision record in the EVO-1860 description.

## Known limitation (deliberate)
Row selection is scoped to the visible page (cleared on page / per-page / search / sort change) to avoid cross-page selection corruption in the shared `BaseTable` (which rebuilds selection from the array it is handed). Bulk actions operate per page. Cross-page selection would be a future enhancement.

## Security
No new endpoints or data exposure; frontend-only; permission gating unchanged.

## Test plan
- `pnpm exec eslint <changed files>` — clean (0 new; pre-existing `any` debt in cannedResponsesService left untouched)
- `pnpm exec vitest related --run` — 0 new failures (`SendCannedResponsePanel` passes → the `getCannedResponses(searchKey?, perPage?)` signature change is backward-compatible; the 4 `Chat.spec` failures are pre-existing on `develop`, proven via stash)
- Manual: Settings → Labels / Canned Responses / Custom Attributes — per-page selector works, page change re-slices, column sort reorders, search filters live (Labels no longer loses items on backspace), no-results state shows "Clear search".

## Changed Files
- `src/pages/Customer/Settings/{Labels,CannedResponses,CustomAttributes}/*.tsx`
- `src/components/{labels,cannedResponses,customAttributes}/*Pagination.tsx`
- `src/services/cannedResponses/cannedResponsesService.ts`
- `src/constants/pagination.ts`
- `src/i18n/locales/{en,pt-BR,es,fr,it,pt}/common.json`

## Linked Issue
- EVO-1860

https://claude.ai/code/session_01Dk9VSPAVSDo3cG8caUbNND

### DIFF
--- a/src/components/cannedResponses/CannedResponsesPagination.tsx
+++ b/src/components/cannedResponses/CannedResponsesPagination.tsx
@@ -16,6 +16,8 @@ export default function CannedResponsesPagination({
   totalCount,
   perPage,
   onPageChange,
+  onPerPageChange,
+  loading,
 }: CannedResponsesPaginationProps) {
   return (
     <BasePagination
@@ -24,6 +26,8 @@ export default function CannedResponsesPagination({
       totalItems={totalCount}
       itemsPerPage={perPage}
       onPageChange={onPageChange}
+      onItemsPerPageChange={onPerPageChange}
+      disabled={loading}
     />
   );
 }
--- a/src/components/customAttributes/CustomAttributesPagination.tsx
+++ b/src/components/customAttributes/CustomAttributesPagination.tsx
@@ -14,8 +14,10 @@ export default function CustomAttributesPagination({
   currentPage,
   totalPages,
   onPageChange,
+  onPerPageChange,
   totalCount,
   perPage,
+  loading,
 }: CustomAttributesPaginationProps) {
   return (
     <BasePagination
@@ -24,6 +26,8 @@ export default function CustomAttributesPagination({
       totalItems={totalCount}
       itemsPerPage={perPage}
       onPageChange={onPageChange}
+      onItemsPerPageChange={onPerPageChange}
+      disabled={loading}
     />
   );
 }
--- a/src/components/labels/LabelsPagination.tsx
+++ b/src/components/labels/LabelsPagination.tsx
@@ -7,14 +7,17 @@ interface LabelsPaginationProps {
   perPage: number;
   onPageChange: (page: number) => void;
   onPerPageChange: (perPage: number) => void;
+  loading?: boolean;
 }
 
 export default function LabelsPagination({
   currentPage,
   totalPages,
   onPageChange,
+  onPerPageChange,
   totalCount,
   perPage,
+  loading,
 }: LabelsPaginationProps) {
   return (
     <BasePagination
@@ -23,6 +26,8 @@ export default function LabelsPagination({
       totalItems={totalCount}
       itemsPerPage={perPage}
       onPageChange={onPageChange}
+      onItemsPerPageChange={onPerPageChange}
+      disabled={loading}
     />
   );
 }
--- a/src/constants/pagination.ts
+++ b/src/constants/pagination.ts
@@ -4,3 +4,8 @@
  */
 export const DEFAULT_PAGE_SIZE = 20;
 export const DEFAULT_PAGE_SIZE_OPTIONS = [10, 20, 50, 100];
+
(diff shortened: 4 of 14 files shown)


dpaes (approved):
## Code Review — EVO-1860 · Approved ✅

Reviewed the full diff against the three acceptance criteria, cross-checking the frontend at the PR head **and** the CRM backend on `develop`. No blockers.

### Acceptance Criteria

**AC1 — accounts with >20 items can reach / search / sort the full catalog (up to the 500 soft-cap)** ✅
The decisive dependency was whether the backend honors `per_page=500`. Confirmed on `develop`: `Api::BaseController#apply_pagination` reads `params[:per_page]` and applies only a **lower** bound (`[page_size, 1].max`) — there is **no upper cap**. `paginate_instance_variables` / `paginated_response` (`api_response_helper.rb`) likewise enforce only a floor. Both `labels#index` and `canned_responses#index` call `apply_pagination`, so the previous silent 20-row cap was the default, not a hard limit — sending `per_page` resolves it. Labels uses `getLabels({ per_page: 500 })`, Canned uses `getCannedResponses(undefined, 500)`; both signature changes are backward-compatible.

**AC2 — page-change / per-page / sort actually re-slice + reorder** ✅
All three screens derive `filtered → sorted → sliced` from state on every render, so page / per-page / sort recompute the slice and ordering. The **items-per-page selector now renders**: the wrappers now forward `onItemsPerPageChange={onPerPageChange}` (previously dropped — the root cause of the dead control). `BasePagination` only mounts the selector when `onItemsPerPageChange` is present — verified.

**AC3 — no dead/cosmetic pager controls remain** ✅
Per-page selector, page buttons and sortable headers are all functional now. The destructive Labels search that overwrote `state.labels` (losing items on backspace) was removed.

### Additional checks (all clean)
- **TypeScript:** all three pagination wrappers declare `onPerPageChange`/`loading`; parents pass consistent props. No type holes (relevant since frontend CI is Sourcery-only and does not run `tsc`).
- **i18n:** `base.noResults.{title,description,clearSearch}` added across all **6 locales** (en, es, fr, it, pt-BR, pt) — parity preserved, `i18n-parity.spec` won't break. `common:` namespace resolves.
- **Page clamping:** `currentPage = Math.min(state.page, totalPages)` prevents empty-page renders; every mutator resets page → 1; deletes reload. No stuck page.
- **Empty state:** correctly distinguishes "no search results" from "create your first".

### Non-blocking follow-ups (optional)
- **[low]** Silent truncation at the 500 cap with no UI hint — the front now ignores the backend `meta.pagination.total`, so a catalog >500 would show "500 of 500" with no indication more exist. Accepted by the card as a documented soft-cap; a defensive `if (meta.total > data.length) warn(...)` would be the natural guard.
- **[low]** No unit tests for the new slice/sort/clamp logic (consistent with repo norms; FE tests not enforced in CI).
- **[info]** Row selection is scoped to the visible page (deliberate M2 tradeoff to avoid cross-page selection corruption in `BaseTable`); bulk actions operate per page.
- **[info]** Header `totalCount` now reflects the filtered count during search (behavior change on Labels) — arguably better UX.

Approving and squash-merging into `develop`.

PR #183, opened 2026-06-23
feat(frontend): canned response attachments UX — show, remove, single-cap (EVO-1861)

nickoliveira23:
## Summary
With the serializer now returning attachments (paired CRM PR), wire the read + remove UX for canned response attachments, cap at a single attachment, and keep the chat composer's attachment sources mutually exclusive.

- Modal: existing attachments render with a remove (X) button; removed ids are tracked, hidden, and sent on save (remove + re-upload = replace). One attachment per canned response — the file input is single, the add button/input disable once one exists, and extra selections are rejected with a toast.
- Service: `CannedResponseFormData.removeAttachmentIds`; `updateCannedResponse` sends `remove_attachment_ids[]` (switches to multipart on adds OR removals).
- Settings table: paperclip + count indicator when a canned response has an attachment.
- Composer: selecting a canned response that has media clears any manually-added files (with a toast), mirroring the existing guard that disables manual upload while canned media is active — so the two sources are never combined.
- Chat dropdown indicator already reads `attachments` and now lights up because the read payload includes them.

## Security
- Frontend-only change; no new data exposure. Server-side size validation lives in the paired CRM PR.

## Test plan
- `pnpm exec tsc -b --noEmit` — clean
- `pnpm exec eslint <changed files>` — 0 errors (pre-existing `any` in the service and a pre-existing `useCallback` warning untouched)
- `pnpm exec vitest related --run` — 0 new failures (`SendCannedResponsePanel` passes; the 4 `Chat.spec` failures are pre-existing on `develop`)
- Manual smoke: upload → reopen shows the attachment + X; remove persists; second attachment blocked; >10 MB blocked; manual-attachment-then-`/`-canned-with-media clears the manual file.

## Changed Files
- `src/components/cannedResponses/CannedResponseModal.tsx`
- `src/components/cannedResponses/CannedResponsesTable.tsx`
- `src/services/cannedResponses/cannedResponsesService.ts`
- `src/components/chat/message-input/MessageInput.tsx`
- `src/types/knowledge/index.ts`

## Related PRs
- crm: https://github.com/evolution-foundation/evo-ai-crm-community/pull/162

## Linked Issue
- EVO-1861

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Dk9VSPAVSDo3cG8caUbNND

## Summary by Sourcery

Add support for managing single attachments on canned responses, including editing, display, and composer behavior, while ensuring attachment changes are correctly sent to the backend.

New Features:
[...]

PR #187, opened 2026-06-23
feat(frontend): full-width composer input + mobile back deselects conversation (EVO-1884)

nickoliveira23:
## Summary
Two chat-conversation UX fixes (found during EVO-1869 mobile smoke):

- **Composer layout** — the message input grew into a large rich-text editor but the composer kept it in a single horizontal row sandwiched between the action buttons, so on narrow widths (mobile, and desktop at smaller sizes) the editor's formatting toolbar was clipped and the text wrapped into an unreadable column. The input now takes the **full composer width** on its own row, and the action buttons move to a **row below** it (attach / emoji / canned / template on the left, mic / send on the right). The Reply/Note toggle + signature/AI row on top is unchanged.
- **Mobile back-button** — the conversation-view back button (`md:hidden`, mobile-only) now clears the selection (reuses `handleCloseConversation` → `clearSelectionAndGoToList`) instead of only toggling the mobile view. Previously `selectedConversationId` stayed set, so re-tapping the same conversation was a no-op (the select handler early-returns on a stale selection). Desktop is unaffected (the back button is hidden on `md+`).

## Security
- Frontend-only; no logic/data/auth surface beyond the chat composer + navigation.

## Test plan
- `pnpm exec eslint src/components/chat/message-input/MessageInput.tsx src/pages/Customer/Chat/Chat.tsx` — 0 errors (a pre-existing `useCallback` deps warning is untouched)
- `pnpm exec vitest related --run` — 0 new failures (the 4 `Chat.spec` failures are pre-existing on `develop`)
- Manual smoke (desktop + mobile ~375px): input full-width with the formatting toolbar fully visible; action buttons in the row below; all buttons functional (attach / emoji / `/`-canned / template / mic / send); mobile back returns to the list and the same conversation can be re-opened.

## Changed Files
- `src/components/chat/message-input/MessageInput.tsx`
- `src/pages/Customer/Chat/Chat.tsx`

## Note
- `MessageInput.tsx` is also touched by EVO-1861 (PR #183, In Review) in a different region (`handleSelectCannedResponse` logic). If #183 merges first, a trivial rebase here may be needed.

## Linked Issue
- EVO-1884

https://claude.ai/code/session_01Dk9VSPAVSDo3cG8caUbNND

### DIFF
--- a/src/components/chat/message-input/MessageInput.tsx
+++ b/src/components/chat/message-input/MessageInput.tsx
@@ -702,62 +702,10 @@ const MessageInput: React.FC<MessageInputProps> = ({
             </div>
           </div>
 
-          {/* Segunda linha: Botões de formatação + Input + Botões de envio */}
-          <div className="flex items-end gap-2 w-full overflow-visible">
-            {/* Botões de formatação à esquerda */}
-            <div className="flex-shrink-0 flex items-center gap-1.5 pb-1">
-              {/* File Upload Button */}
-              <FileUpload
-                onFilesSelected={handleFilesSelected}
-                maxFileSize={100}
-                multiple={true}
-                disabled={isDisabled || isSending || isPendingConversation || hasCannedMedia}
-              />
-
-              {/* Emoji Button */}
-              <div className="relative">
-                <Button
-                  variant="ghost"
-                  size="icon"
-                  disabled={isDisabled || isSending || isPendingConversation}
-                  className="h-9 w-9 flex-shrink-0 hover:bg-accent disabled:opacity-50"
-                  onClick={handleEmojiClick}
-                >
-                  <Smile className="h-4 w-4" />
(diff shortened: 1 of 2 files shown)


dpaes (approved):
Reviewed against EVO-1884. Both fixes are correct and minimal: the composer input goes full-width with the action buttons relocated to a `justify-between` row below it (formatting toolbar no longer clipped, input no longer crushed on mobile), and the mobile back button now routes through `clearSelectionAndGoToList` which dispatches `selectConversation(null)` — genuinely clearing `selectedConversationId` so the same conversation can be re-tapped to reopen. No desktop regression. Heads-up: #183 (EVO-1861) also touches `MessageInput.tsx` in a different region and may need a trivial rebase if it merges first — not a conflict today. LGTM — approving.

PR #189, opened 2026-06-24
feat(journey): warn on no-exit cycles in pre-activation validation (EVO-1857)

daniloleonecarneiro:
## Summary
Follow-up to EVO-1744 (adversarial finding F5). `validateJourneyTerminalPaths` (the EVO-1692 seed, folded into the `validateJourney` engine) flags only nodes with **no outgoing edge**, so a closed loop with no reachable exit/transfer node (`trigger → A → B → A`) **passed** validation — yet at runtime the contact's session stays "running" for ~30 days (EVO-1691). Every cyclic node has an outgoing edge, so the existing BFS never sees it (`journeyFlowValidation.ts:39` literally noted "Cyclic paths with no exit are not detected").

This adds an `unreachableExit` rule (warning, same severity as the dangling check) to the engine:
- One **reverse-reachability** pass from the terminal nodes (`exit-journey-node`/`transfer-journey-node`) yields the set that *can* reach a terminal; any trigger-reachable node outside it is trapped → warn.
- Excludes trigger nodes (the entry, not the offending step) and no-outgoing nodes (already `terminalPath`'s `danglingExit` domain) so it never double-flags.
- Linear (O(V+E)). Consumers render `t(issue.messageKey)` generically, so the new rule surfaces in the editor banner, node badge, and list/modal with no further wiring.

## Test plan
- `vitest run src/utils/journeyFlowValidation.spec.ts` — 15 tests (+4): cycle-without-exit warns (AC1), cycle-with-reachable-exit is **not** a false positive (AC2), self-loop warns, and a detached (non-trigger-reachable) cycle is ignored.
- Validation-related suites green (`journeyFlowValidation` + `journeyValidators`, 29 tests). `journeyFlowValidation.spec.ts` is the only spec that imports the engine.
- `tsc -b` clean (repo convention). All 6 locale JSONs valid.

## Notas
- Adds `flowEditor.validation.unreachableExit` in en/es/fr/it/pt/pt-BR.
- Pure FE engine change; no runtime (evo-flow) changes. Builds on `develop` which has EVO-1744 (#180).

### DIFF
--- a/src/i18n/locales/en/journey.json
+++ b/src/i18n/locales/en/journey.json
@@ -92,6 +92,7 @@
       "requiredConfig": "{{node}}: missing required configuration ({{fields}})",
       "triggerActionContext": "{{node}}: needs a conversation, but the trigger doesn't provide one",
       "danglingExit": "{{node}}: this path doesn't end in an exit or transfer node",
+      "unreachableExit": "{{node}}: this node is stuck in a loop that can never reach an exit or transfer node",
       "summaryErrors": "Fix these before activating:",
       "summaryWarnings": "Warnings:",
       "saveIssues": "Saved with pending issues: {{issues}}",
--- a/src/i18n/locales/es/journey.json
+++ b/src/i18n/locales/es/journey.json
@@ -92,6 +92,7 @@
       "requiredConfig": "{{node}}: falta configuración obligatoria ({{fields}})",
       "triggerActionContext": "{{node}}: necesita una conversación, pero el disparador no la proporciona",
       "danglingExit": "{{node}}: esta ruta no termina en un nodo de salida o transferencia",
+      "unreachableExit": "{{node}}: este nodo está atrapado en un bucle que nunca puede llegar a un nodo de salida o transferencia",
       "summaryErrors": "Corrige esto antes de activar:",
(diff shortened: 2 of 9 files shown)


dpaes:
**Approved — EVO-1857.**

The no-exit-cycle detection is correct: reverse-BFS from terminal nodes (`exit`/`transfer`) intersected with forward-reachability from triggers, emitting `unreachableExit` for trigger-reachable nodes that can't reach a terminal — with exclusions for trigger nodes and zero-outgoing nodes (the dangling domain) so it never double-flags. The AC2 false-positive trap (a detached cycle with no trigger) is handled because emission iterates only over trigger-reachable nodes. All 3 ACs met; the `unreachableExit` i18n key is present across the 6 locales; integrates into the 3 render surfaces via the generic `t(issue.messageKey)` with no extra wiring.

One non-blocking copy nit, split into a follow-up (EVO-1889): in a *linear* dangling chain (`trigger → a → b`, where `b` has no exit), node `a` gets `unreachableExit` whose message says "stuck in a loop" — but `a` isn't in a cycle. Tracked separately for a copy/scope tweak; doesn't block this merge.

Note: frontend CI does not run vitest/tsc — tests are author-reported. Merging (squash) to develop.

PR #167, opened 2026-06-17
fix(frontend): emit segment trigger action in journey flowTriggers (EVO-1763)

nickoliveira23:
## Summary
The journey editor dropped `segmentAction` (and `segmentName`) when building `flowTriggers`, so the evo-flow segment matcher never received it and couldn't honor entered/exited. (`labelAction` was already sent.)

- Extracted the inline builder into `buildFlowTriggers(nodes)` (pure, unit-tested) and mirror `segmentId`/`segmentName`/`segmentAction` into the trigger metadata.
- This is the companion to the evo-flow matcher fix — without it, segment triggers always default to `entered`.

## Migration note (segment journeys)
Segment journeys saved **before** this change have no `segmentAction` persisted, so the runtime defaults them to `entered`. A journey intended as **"exited"** must be **re-opened and saved once** to start matching exited events again. Label journeys are unaffected.

## Validation
- `evo-ai-frontend-community: pnpm exec vitest run journeyFlowTriggers.spec.ts` → 4/4
- `evo-ai-frontend-community: pnpm exec tsc -b --noEmit` → clean
- `evo-ai-frontend-community: pnpm exec eslint <new files>` → clean
- **No live smoke:** saving a journey + trigger matching is evo-flow-backed end to end; the stack (Kafka/Temporal/evo-flow) was not up locally. Behavior covered by unit tests here + in the evo-flow PR. Live smoke pending in a full environment.

## Changed Files
- `src/pages/Customer/Journey/JourneyFlowEditor.tsx`
- `src/pages/Customer/Journey/journeyFlowTriggers.ts` (new)
- `src/pages/Customer/Journey/journeyFlowTriggers.spec.ts` (new)

## Related PRs
- evo-flow: https://github.com/evolution-foundation/evo-flow/pull/70

## Linked Issue
- EVO-1763

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Extract flow trigger construction into a reusable helper and ensure segment trigger metadata is persisted for evo-flow matching.

Bug Fixes:
- Persist segmentId, segmentName, and segmentAction in journey flow trigger metadata so evo-flow can correctly honor segment trigger actions.

Enhancements:
- Extract journey trigger-node processing into a pure buildFlowTriggers utility reused by the journey editor.

Tests:
- Add unit tests for buildFlowTriggers to cover trigger payload and metadata construction.

PR #192, opened 2026-06-24
feat(frontend): WhatsApp delete/revoke UX — keep content + notices (EVO-1890, EVO-1891)

nickoliveira23:
Frontend for **EVO-1890** (inbound contact-revoke notice) + **EVO-1891** (outbound delete propagation). Pairs with the CRM PR #167.

## Summary
Deleted messages now **keep their content** and show a notice, instead of hiding behind "This message was deleted":
- **EVO-1890:** when `content_attributes.revoked_by_contact` is set (contact deleted for everyone), the bubble shows the original content **plus** a "Deleted by the contact" notice.
- **EVO-1891:** when an agent-deleted message could not propagate to the provider (`content_attributes.revoke_propagated === false`), a "removed only in the CRM" hint is shown next to the notice.
- Agent-deleted messages (the existing `deleted` flag) also keep their content now (product decision) + show the "deleted" notice.

`MessageBubble` updates both render paths (regular + facebook-style); i18n keys `revokedByContact` / `revokeCrmOnly` across all 6 locales; `Message.content_attributes` type gains `revoked_by_contact` / `revoke_propagated`.

## Test plan
- `evo-ai-frontend-community: pnpm exec tsc -b --noEmit` (clean)
- locale JSON valid (6/6)
- Manual smoke (evolution provider): contact revoke → content + "Apagada pelo contato"; agent delete → content + notice.

## Changed Files
- `src/components/chat/messages/MessageBubble.tsx`
- `src/types/notifications/messages.ts`
- `src/i18n/locales/{en,pt-BR,pt,es,fr,it}/chat.json`

## Related PRs
- crm: https://github.com/evolution-foundation/evo-ai-crm-community/pull/167

## Linked Issues
- EVO-1890, EVO-1891

## Summary by Sourcery

Update WhatsApp message bubbles to retain deleted message content while adding contextual deletion notices for agents and contacts.

New Features:
- Display a notice when a message is revoked by the contact while keeping the original message content visible.
- Indicate when an agent-deleted message was removed only in the CRM but not propagated to the provider.

Enhancements:
- Always render message content in bubbles and append deletion or revocation notices instead of replacing the content.
- Extend message content attributes with revoke metadata flags to support new deletion states.
- Add i18n strings for contact revocation and CRM-only deletion notices across supported locales.

### DIFF
--- a/src/components/chat/messages/MessageBubble.tsx
+++ b/src/components/chat/messages/MessageBubble.tsx
@@ -77,6 +77,8 @@ const MessageBubble: React.FC<MessageBubbleProps> = ({
   const isActivity = message.message_type === MESSAGE_TYPE.ACTIVITY;
   const isPrivate = message.private;
   const isDeleted = message.content_attributes?.deleted === true;
+  const isRevokedByContact = message.content_attributes?.revoked_by_contact === true;
+  const isRevokeCrmOnly = message.content_attributes?.revoke_propagated === false;
 
   // Moderation indicators
   const hasPendingModeration = moderation?.status === 'pending';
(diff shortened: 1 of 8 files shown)


dpaes:
## Note (EVO-1890/1891) — AC1 gated on crm#167; one product flag

The FE here looks fine on its own; I'm not blocking it for a code defect. Two notes:
1. **AC1** (the "deleted by contact" notice actually showing) is gated on crm#167 — the backend revoke mark isn't proven to surface yet (id-format risk per the card's own smoke note). Once that's verified, please confirm the notice render here end-to-end.
2. **Product decision (non-blocking, not a code defect):** this PR also changes AGENT-deleted messages to show their content + a notice instead of hiding it (`MessageBubble`: `isDeleted ? placeholder : content` → always content). That's a UX call outside the card scope — worth a quick confirm with product before it ships.

dpaes (approved):
## Re-review — Approved (AC1 addressed)

The AC1 reject (the "deleted by contact" notice not proven to surface) is resolved. New spec `spec/services/whatsapp/evolution_revoke_integration_spec.rb` proves the flow end-to-end **without stubbing `find_by`**:
- creates a real `Message` with `source_id`, runs the real `messages.delete` event through `IncomingMessageEvolutionService`, reloads and asserts `revoked_by_contact == true` + content preserved;
- **test 3 (decisive):** ingests via the real `messages.upsert`, then revokes by the **same key id**, proving `inbound source_id == delete id` — the exact id-format concern from the earlier smoke;
- incoming-only guard (fromMe) + no-match no-op, both with real persistence.

Caveat: couldn't run it locally this session (Docker engine offline), so this rests on the test design + author's run + the code path I read (`revoked_message_source_id` reads `key.id`; inbound stores `source_id = key.id`). CRM CI runs no rspec.

Non-blocking follow-ups (as previously agreed): the product flag (agent-deleted content show/hide, fe#192) and the Z-API `delete_message` arity (EVO-1891).

PR #83, opened 2026-05-18
fix(dev): relax nginx CSP in development to allow localhost and widget embedding

miltonsosa-info-unlp:
## Summary
- Enable local development without CSP blocks between the frontend and services.
- In [docker-entrypoint.sh](cci:7://file:///home/miltonsosa/trabajo/evo-crm-community/evo-ai-frontend-community/docker-entrypoint.sh:0:0-0:0), when `VITE_APP_ENV=development` (default), relax nginx CSP to:
  - Add `http://localhost:*` to `connect-src`
  - Set `frame-ancestors *` so the widget can be embedded during dev
- No changes for production; CSP remains strict.

## Root cause
- In local dev, the app performs API/WebSocket calls to `http://localhost:<port>`. Default CSP (`connect-src 'self' https: wss: ws:`) blocks those HTTP calls to localhost.
- Widget embedding/preview in dev requires permissive `frame-ancestors`; otherwise the iframe is blocked.

## Changes
- File: [evo-ai-frontend-community/docker-entrypoint.sh](cci:7://file:///home/miltonsosa/trabajo/evo-crm-community/evo-ai-frontend-community/docker-entrypoint.sh:0:0-0:0)
  - Determine env: `APP_ENV="${VITE_APP_ENV:-development}"`
  - If development, patch nginx config at runtime:
    ```sh
    sed -i "s|connect-src 'self' https: wss: ws:|connect-src 'self' https: wss: ws: http://localhost:*|g" /etc/nginx/conf.d/default.conf
    sed -i "s|frame-ancestors 'self'|frame-ancestors *|g" /etc/nginx/conf.d/default.conf
    ```

## Test checklist
- Start container with default env (or `VITE_APP_ENV=development`):
  - Confirm network requests to `http://localhost:<ports>` succeed (no CSP violations).
  - Check response headers show CSP with `connect-src ... http://localhost:*`.
  - If embedding the widget locally, confirm it loads within an iframe.
- Start with `VITE_APP_ENV=production`:
  - Confirm CSP remains strict (no `http://localhost:*`, `frame-ancestors 'self'`).
  - App serves as usual.

## Backward compatibility
- Dev-only behavior by default. Production unchanged unless `VITE_APP_ENV=production` is explicitly set.
- No app code changes; only nginx headers modified at container startup.

## Risks and mitigations
[...]

PR #194, opened 2026-06-25
fix(journey): conditional handle prefix (EVO-1902)

daniloleonecarneiro:
## Summary

Reworked após review. A versão anterior trocava o id do Handle do nó Conditional de `path-<id>` para o `path.id` cru — o que **regredia jornadas legadas**: as edges salvas têm `sourceHandle = 'path-<id>'`, então contra um handle cru `isHandleConnected()` as renderizava cinza/desconectadas e não se auto-curavam no save.

A decisão canônica é manter `path-<id>` no Handle:

- **Render**: `Handle.id === edge.sourceHandle === 'path-<id>'` tanto para jornadas legadas quanto novas → handle conectado (verde), sem regressão.
- **Roteamento**: já resolvido em runtime pela **EVO-1922 / #99** (já mergeada em develop), que normaliza o prefixo `path-` nos **dois lados** da comparação (`normalizeConditionalPathHandle`). Manter `path-<id>` no FE casa com o `path.id` cru do backend após a normalização **sem double-strip** (a EVO-1922 faz um único strip).
- Consistente com o nó **Split**, que também prefixa nos dois lados.

Não é necessária normalização em load-time: como a forma canônica `path-<id>` já mantém `Handle.id === edge.sourceHandle` para edges legadas e novas, não existem edges fora do padrão para curar.

## Files
- `src/components/journey/nodes/actions/conditional/ConditionalNode.tsx` — reverte o handle para `path-${path.id}` + comentário explicando a relação com a EVO-1922.
- `src/components/journey/nodes/actions/conditional/ConditionalNode.spec.tsx` — novo teste.

## Test plan
- `npx tsc -b` → OK (exit 0)
- `npx vitest run src/components/journey/nodes/actions/conditional/ConditionalNode.spec.tsx` → 3/3 verdes:
  - edge legada `sourceHandle='path-<id>'` renderiza conectada (verde)
  - sem edge → desconectada (cinza)
  - edge crua `sourceHandle='<id>'` NÃO conecta (guarda contra reintroduzir a mudança rejeitada)

## AC
Jornada condicional criada no editor roteia o ramo true corretamente quando a condição é satisfeita (EVO-1922 runtime), e jornadas legadas renderizam o ramo como conectado (sem regressão de cinza/desconectado).

### DIFF
--- a/src/components/journey/nodes/actions/conditional/ConditionalNode.spec.tsx
+++ b/src/components/journey/nodes/actions/conditional/ConditionalNode.spec.tsx
@@ -1,0 +1,75 @@
+import { render } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { ReactFlowProvider, type Edge } from '@xyflow/react';
+import { ConditionalNode, type ConditionalNodeData } from './ConditionalNode';
+import '@/i18n/config';
+
+// The node reads connected handles via useEdges(); mock it so we can feed the
+// exact edges a saved (legacy) journey would hydrate into the editor.
+const edgesMock = vi.hoisted(() => ({ current: [] as Edge[] }));
+
+vi.mock('@xyflow/react', async importOriginal => {
+  const actual = await importOriginal<typeof import('@xyflow/react')>();
+  return {
+    ...actual,
+    useEdges: () => edgesMock.current,
+  };
+});
+
+function renderNode(data: ConditionalNodeData, edges: Edge[]) {
+  edgesMock.current = edges;
+  return render(
+    <ReactFlowProvider>
+      <ConditionalNode id="cond-1" selected={false} data={data} />
+    </ReactFlowProvider>,
+  );
+}
+
+const data: ConditionalNodeData = {
+  label: 'Conditional',
+  paths: [
+    {
+      id: 'p1',
+      name: 'True branch',
+      logicalOperator: 'AND',
+      conditions: [{ id: 'c1', type: 'contact', field: 'contact.name', operator: 'equals', value: 'x' }],
+    },
+  ],
+};
+
+// EVO-1902: the Handle id stays `path-<id>` so it matches the sourceHandle that
+// legacy journeys already saved. Routing is normalized at runtime by EVO-1922.
+describe('ConditionalNode — handle connectivity (EVO-1902)', () => {
+  it('renders the TRUE branch handle as connected for a legacy edge sourceHandle="path-<id>"', () => {
+    const { container } = renderNode(data, [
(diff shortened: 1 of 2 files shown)


dpaes (changes requested):
## Changes requested — EVO-1902

EVO-1902's acceptance criterion is already satisfied by **EVO-1922 / #99** (approved), which normalizes an optional `path-` prefix on *both* sides of the conditional match at runtime — covering legacy AND new journeys without the FE needing to emit the raw id. Given that, this FE change is redundant and introduces a regression:

1. **Redundant for runtime** — #99 already routes both `path-<id>` and `<id>` correctly.
2. **Cosmetic editor regression (legacy journeys).** Changing the Handle id from `path-${path.id}` to raw `path.id` makes `isHandleConnected()` (`edge.sourceHandle === handleId`) return false for journeys saved before this change (persisted `sourceHandle='path-<id>'` vs new handle id `path.id`) → the TRUE-path handle renders grey/"not connected". No data loss and no routing break (with #99), but it's persistent and does **not** self-heal on re-save (the sourceHandle stays `path-<id>` verbatim).
3. **Inconsistency** — leaves Conditional emitting a raw id while Split keeps its `split-variant-<id>` prefix on both sides.

**What's missing / path forward:** since #99 already closes the AC, the cleanest outcome is to **drop/close this PR** (EVO-1902 resolved by #99 — same pattern as EVO-1892↔EVO-1929). If you prefer to keep the FE canonical, this PR must also **normalize `edge.sourceHandle` on load** (rewrite `path-<id>`→`path.id`) so legacy journeys render connected and the change self-heals on save.

dpaes (approved):
Macro review — approved.

The FE correctly stays on `path-${path.id}`; the actual D13 routing fix is carried by EVO-1922/#99 (already on develop), which normalizes the `path-` prefix on both sides (normalizeConditionalPathHandle). This PR is a no-op on production code plus a regression guard (ConditionalNode.spec.tsx) against re-introducing the rejected raw-id change. This resolves the round-1 review: keeping `path-<id>` avoids the legacy-edge gray-out regression and makes load-time normalization unnecessary.

Note (non-blocking): the spec covers editor handle connectivity, not the routing AC itself — that behavior lives in evo-flow/#99 and isn't testable in this repo. Branch CLEAN on develop. Squash-merging.

miltonsosa-info-unlp:
@daniloleonecarneiro &  @dpaes  Please review the references in this PR description, as they appear to be incorrect.

PR #99   (fix(dev): relax nginx CSP in development to allow localhost and widget embedding) is a pending PR I submitted that fixes CSP validation in docker-entrypoint.sh and updates vitest. It is a follow-up to my earlier PR #83.

It has nothing to do with normalizeConditionalPathHandle, path- prefix normalization, or journey conditional node routing.

PR #214, opened 2026-06-26
fix(frontend): reset global filters on "clear filters" so the badge clears (EVO-1939)

nickoliveira23:
## Summary

On the conversations sidebar, clicking **"Clear filters"** reloaded open conversations but the filter **badge kept showing the stale count** — the filter looked like it was never removed (symptom B, reported by a builder's customer).

Root cause: `handleClearFilters` (`useFilterHandlers.ts`) only cleared local component state + localStorage; it never reset the **global** `FiltersContext.state.activeFilters`. The badge (`ChatSidebar`) reads that global state, and the sync effect even re-mirrored the old filters back into the modal.

## Fix

`handleClearFilters` now resets the global filters to the default open filter (`[DEFAULT_FILTER]`).

**Why reset to the default and not to `[]`:** the realtime conversation matcher (`doesConversationMatchFilters`, `ChatContext`) treats an empty filter list as **match-all** — resetting to `[]` would let non-open conversations leak into the list on websocket updates. Keeping `[status=open]` preserves the existing open-only behavior everywhere (matcher, search re-scoping, reload). The only behavior that changes is the badge.

The badge now counts only **non-default** filters, so it disappears at the default state (and no longer shows a spurious "1 filter" on a fresh load).

## Changes

- `src/contexts/chat/FiltersContext.tsx` — export `DEFAULT_FILTER`
- `src/hooks/chat/useFilterHandlers.ts` — reset global filters in `handleClearFilters`
- `src/components/chat/chat-sidebar/ChatSidebar.tsx` — badge counts only non-default filters
- `src/hooks/chat/__tests__/useFilterHandlers.spec.ts` — covers the global reset on clear

## Validation

- `frontend: pnpm exec tsc -b --noEmit` → 0 errors
- `frontend: pnpm exec eslint` (changed files) → 0 errors (2 pre-existing react-refresh warnings on the context file)
- `frontend: pnpm exec vitest run` (useFilterHandlers + ChatSidebar specs) → 18 passed
- **Manual smoke:** apply → clear hides the badge, shows open conversations, modal resets to default; F5 does not restore old filters; applying a real filter still works; realtime stays scoped to open.

## Acceptance Criteria

- [x] AC1 (scroll keeps the filter) — already covered by EVO-1671 (loadMore preserves the query) on develop
- [x] AC2 — "Clear filters" resets `activeFilters` to default, badge disappears immediately, list shows open conversations
- [x] AC3 — after clear + F5, no previous filter is restored
- [x] AC4 — vitest covers the apply→clear reset

## Linked Issue

- EVO-1939

https://claude.ai/code/session_01NzVxXUXPKJQy4m5WuJEjgz

### DIFF
--- a/src/components/chat/chat-sidebar/ChatSidebar.tsx
+++ b/src/components/chat/chat-sidebar/ChatSidebar.tsx
@@ -554,6 +554,18 @@ const ChatSidebar = ({
   const totalPages = pagination?.total_pages || 1;
   const hasNextPage = pagination?.has_next_page ?? currentPage < totalPages;
 
+  // EVO-1939: o filtro padrão (status=open) representa a visão default e não
+  // deve contar como filtro aplicado no badge — assim o badge some ao limpar.
+  const appliedFilterCount = filters.state.activeFilters.filter(
+    f =>
+      !(
+        f.attribute_key === 'status' &&
+        f.filter_operator === 'equal_to' &&
+        f.values?.length === 1 &&
+        f.values[0] === 'open'
+      ),
+  ).length;
+
   const handleSidebarScroll = useCallback(async () => {
     const now = Date.now();
     if (now - lastScrollTimeRef.current < 150) return;
@@ -992,11 +1004,11 @@ const ChatSidebar = ({
               : t('chatSidebar.conversations')}
           </span>
           <div className="flex items-center gap-2">
-            {/* Indicador de filtros ativos */}
-            {filters.state.activeFilters.length > 0 && (
+            {/* Indicador de filtros ativos (ignora o filtro padrão status=open) */}
+            {appliedFilterCount > 0 && (
               <Badge variant="secondary" className="text-xs">
-                {filters.state.activeFilters.length}{' '}
-                {filters.state.activeFilters.length === 1
+                {appliedFilterCount}{' '}
+                {appliedFilterCount === 1
                   ? t('chatSidebar.filter')
                   : t('chatSidebar.filters')}
               </Badge>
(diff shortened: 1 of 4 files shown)


dpaes (approved):
Reviewed against the card + full diff. **Approving.**

**Scope — closes EVO-1939.** Symptom B (the only live cause per the card) is fully fixed: `handleClearFilters` now calls `filters.setFilters([DEFAULT_FILTER])` (`useFilterHandlers.ts:44`), resetting the global `FiltersContext` the badge reads. AC2 met (global reset + badge hides via `appliedFilterCount`). AC3 met on both F5 restore paths (`Chat.tsx:185`, `ChatContext.tsx:669`). Reset-to-default (not `[]`) is correctly justified — `[]` is treated as match-all and would leak non-open conversations on websocket updates. Symptom A (scroll) is correctly out of scope (already on develop via EVO-1671; the card itself states A is fixed).

**Follow-ups (low/info, not blocking):**
- AC4 only partially covered: the new spec tests the hook clear-path but not the new badge `appliedFilterCount` logic (`ChatSidebar.tsx:559-568`) nor the apply→scroll flow AC4 lists. Frontend runs no vitest in CI, so this is unverified by automation.
- Badge undercounts when an explicit `status=open` filter is combined with others (the predicate excludes any `{status,equal_to,['open']}`, not just the auto-injected default). Cosmetic.
- The test mocks `DEFAULT_FILTER` + adapter instead of importing the real ones → a future shape drift wouldn't fail the test.
- Three encodings of the `status=open` default now coexist (`DEFAULT_FILTER`, `DEFAULT_CONVERSATION_FILTER`, the hardcoded badge predicate); they agree today but drift in any one silently breaks badge-hide. Pre-existing duplication; the PR adds the third.

PR #50, opened 2026-05-08
feat(chat): bulk resolve conversations via checkbox selection (EVO-1011)

marcelogorutuba:
## Summary
- Replace hardcoded PT-BR toasts with `t()` i18n calls in all 6 locales (`bulkResolveSuccess`, `bulkResolvePartialSuccess`, `bulkResolveError`)
- Consume `success_ids`/`failed_ids` from backend response to show per-item failure toast
- Clear checkbox selection when user opens a conversation row (AC #5)
- Guard `handleBulkResolve` with `can('conversations', 'update')` permission check
- Use `reloadCurrentFilters()` after bulk resolve instead of raw `updateConversation` (fixes resolved conversations staying visible in open filter)
- Cap bulk selection at 200 items client-side
- Wire `chatSidebar.selected` i18n key into toolbar span (was dead)
- Remove `eslint-disable react-hooks/exhaustive-deps` on `onClearSelection` effect — add dep instead
- Add `aria-label` to conversation row `<Checkbox>` for screen readers
- Use `checked` boolean in `onCheckedChange` handler
- Move `bulkResolve` call from `conversationAPI` import to `chatService` facade
- Add vitest spec for `chatService.bulkResolve` (3 cases) and selection cap logic
- Sidebar width stays `w-80` and only expands to `w-96` when bulk toolbar is visible

## Validation
- `evo-ai-frontend-community: pnpm exec tsc -b --noEmit` ✅
- `evo-ai-frontend-community: pnpm lint` (0 errors in changed files) ✅
- `evo-ai-frontend-community: npx vitest run src/services/chat/chatService.bulkResolve.spec.ts` ✅ 4/4 passed

## Changed Files
- `src/pages/Customer/Chat/Chat.tsx`
- `src/components/chat/chat-sidebar/ChatSidebar.tsx`
- `src/services/chat/chatService.ts`
- `src/services/chat/chatService.bulkResolve.spec.ts`
- `src/i18n/locales/en/chat.json`
- `src/i18n/locales/es/chat.json`
- `src/i18n/locales/fr/chat.json`
- `src/i18n/locales/it/chat.json`
- `src/i18n/locales/pt/chat.json`
- `src/i18n/locales/pt-BR/chat.json`

## Related PRs
- Backend: https://github.com/evolution-foundation/evo-ai-crm-community/pull/58

## Linked Issue
- EVO-1011

PR #215, opened 2026-06-26
feat(frontend): make the Users list filter + search functional (EVO-1947 Fase A)

nickoliveira23:
## Summary

EVO-1947 **Fase A (Users)** — frontend half. Makes the advanced-filter UI on the
Users (Team Members) screen actually work, now that the auth backend honors
`filters[]`/`q`/`sort`.

## Changes

- Re-enable the filter button (`showFilters={true}` on `UsersHeader`) — it was
  hidden as a stopgap while the filter was dead.
- `loadUsers` accepts the filters explicitly (`filtersOverride`) so the first
  apply/clear sends the **just-applied** filters instead of the stale
  `activeFilters` closure (AC3).
- The applied-chip **"x"** computes the new filter list from a ref kept in sync
  with `activeFilters`, so removing one chip removes the right filter rather than
  a stale snapshot.
- The **search term** is carried on every reload (ref + debounce + request
  sequence), so paginating or applying a filter no longer re-fetches the
  unsearched list while the box still shows the term.
- The **role options** come from the roles endpoint instead of the hard-coded
  administrator/agent pair, falling back to the built-in pair when listing roles
  is denied.
- The **chosen sort** is carried on every reload too. `loadUsers` hardcoded
  `sort=name&order=asc`, so pagination/search/filter-apply silently reverted the
  ordering while the header kept pointing at the picked column — and page 2 of a
  different ordering repeats and skips rows. `UsersState.sortBy` is narrowed to
  the columns the backend whitelists.
- The **ID column is display-only** (the backend has no `id` sort; it was a dead
  control, and sorting by UUID is meaningless).
- Applied chips render via the shared `buildAppliedFilterChips` helper (AC4).

## Security

- No new data exposure; filtering/search/sort are enforced server-side (auth).

## Test plan

- `npx vitest run src/pages/Customer/Settings/Users` → 10 passed
  (search carried across pagination/filter, debounce, role options + fallback,
  and the sort carried across pagination/filter/search + asc↔desc toggle).
  Control negative: pinning `sort` back to `'name'` fails the 4 sort examples.
- `npx tsc -b --noEmit` → 0 errors
- `npx eslint` on the changed files → no new offence (4 pre-existing errors on
  untouched lines, confirmed against `develop`)
- The spec is on the CI opt-in list, so it runs on every PR.
- Manual smoke: filter button visible; apply changes the list on the first click;
  search filters; chip "x" removes the correct filter; clear resets.

## Changed Files

- `src/pages/Customer/Settings/Users/Users.tsx`
- `src/pages/Customer/Settings/Users/Users.spec.tsx` (new)
- `src/components/users/UsersFilter.tsx` (accepts a `filterTypes` override)
- `src/components/users/UsersTable.tsx` (ID column no longer sortable)
- `src/types/users/users.ts` (`sortBy` narrowed to the sortable columns)
- `.github/workflows/test.yml` (spec added to the opt-in Vitest list)

## Related PRs

- auth (server-side filter/search/sort): `feat/EVO-1947` on
  evo-auth-service-community (#50).

## Linked Issue

- EVO-1947 (Fase A — Users; Agents/CustomTools/MCP landed as EVO-1952/EVO-1953,
  Tools canceled as EVO-1954)

https://claude.ai/code/session_01NzVxXUXPKJQy4m5WuJEjgz


dpaes (dismissed):
Reviewing as a pair (frontend #215 + auth #50). The **Users (Fase A) slice is solid** — backend is SQL-injection-safe (attribute keys whitelisted, values bound, `query_operator` restricted to AND/OR), the base relation is identical to the prior scope (no authorization/data-exposure regression), FE↔BE attribute+operator coverage is exhaustive (no silent no-ops), and AC1/AC3/AC4 are met for Users.

**Holding approval.** EVO-1947 covers 5 list screens across 3 repos, and this pair delivers only Users. Per the team call, approval is conditioned on the card's **full ACs** being completed — i.e. **AC2** (the Agents filter modal, still `onFilterClick={() => {}}`) and the **4 core-service (Go) screens** (Agents / Tools / Custom Tools / Custom MCP Servers), the deferred Fase B/C. Once those land (here or stacked), I'll re-review and approve the set together so the card closes in one move. Leaving this PR open.

**Non-blocking notes for when you pick Fase B/C back up:**
- Neither repo runs tests in CI (auth no rspec, frontend no jest/tsc), so the 17+3 specs + tsc-clean are local-only.
- Mixed AND/OR across 3+ filters relies on SQL precedence: `frag1 AND frag2 OR frag3` → `(frag1 AND frag2) OR frag3` (mirrors `Contacts::FilterService`).
- `handleApplyFilters` drops the active `q` search term from the request (pre-existing asymmetry, not worsened here).
- Old private `users_controller#users` is now dead code (`FilterService#base_relation` replaces it verbatim).

gomessguii (approved):

PR #239, opened 2026-07-06
ci: guard against reintroducing internal-only identifiers

gomessguii:
Adds a CI check that fails if a PR introduces identifiers that must not exist in this repo. Defensive hygiene — no functional change.


DavidsonGomes (approved):
aprovado

PR #340, opened 2026-08-28
feat(phone-input): default the phone country to the active app locale

mateusbellozupko:
## Summary
- The shared `PhoneInput` component's `defaultCountry` was hardcoded to `'BR'` in both the widget's pre-chat form and the admin Contact form, regardless of which locale the app was actually running in.
- Adds `localeToPhoneCountry.ts`, mapping the app's active locale to a sensible default calling-code country: `pt-BR`→BR, `pt`→PT, `en`→US, `es`→ES, `fr`→FR, `it`→IT.
- Both call sites now derive `defaultCountry` from `currentLanguage` instead of the hardcoded value. Users can still manually pick a different country from the flag dropdown — this only changes the initial default.

## Testing notes
- Verified `useLanguage()`'s `currentLanguage` is the global i18n language (not scoped by the `useLanguage(namespace)` argument), so both call sites (`widget` and `contacts` namespaces) read the same value correctly.
- No behavior change when the app locale isn't one of the six currently supported locales — falls back to `BR`.

## Summary by Sourcery

Localize phone country defaults and improve country selection across contact and pre-chat phone fields.

New Features:
- Set phone input defaults according to the active app locale, with a Brazil fallback for unsupported locales.
- Provide a searchable, flag-based country picker for shared phone inputs.
- Render widget pre-chat phone fields with a country selector and localized phone-number formatting.

Bug Fixes:
- Ensure previously saved pre-chat phone fields using the legacy text type continue to receive phone-specific rendering and validation.
- Preserve selected phone countries and correctly compose international phone values when editing or changing countries.

Enhancements:
- Add explicit phone field typing and localized country labels for widget pre-chat forms.

Tests:
- Add coverage for the country combobox interactions, phone value formatting, and legacy/current pre-chat phone field rendering.


mateusbellozupko:
Follow-up: the widget pre-chat phone field now uses a country-code dropdown + plain digits input (`PhoneNumberField`) instead of the combined `react-phone-number-input` field, since that library only applies `defaultCountry` on mount and never corrects itself once the widget's locale resolves asynchronously after the config fetch. `ContactForm.tsx` (admin contact editing) is unaffected — still uses the original shared `PhoneInput` + `localeToPhoneCountry`.

mateusbellozupko:
Added two more fixes found while testing this feature:

- **fix(widget): give the phone pre-chat field type 'phone', not 'text'** — `getDefaultPreChatFields()` defaulted the phoneNumber field's `type` to `'text'`, so `PreChatForm`'s `switch(field.type)` never hit its own `case 'phone':` branches. The field silently rendered as a plain text input with no country dropdown, even once enabled in channel settings. Fixed the default and added a name-based fallback so accounts with an already-saved `type: 'text'` config pick up the fix without re-saving.
- **fix(phone-input): replace the native country `<select>` with a searchable combobox** — the library's default country picker is an unstyleable native `<select>`, which looked out of place (plain OS popup, no flags visible, low contrast). Added `CountrySelectCombobox` (Popover+Command, same pattern as `ModelSelector`/`AssigneeFilter`) and wired it in via `countrySelectComponent`. Fixes it everywhere `PhoneInput` is used: Create Contact and every WhatsApp/SMS channel form.

PR #348, opened 2026-08-31
fix(agents): refresh the pinned model list across every provider (CRM-442)

daniloleonecarneiro:
The Google axis of `availableModels` was swept once; the rest of the pinned list was not, and by now almost none of it exists. Every Anthropic entry was a retired dated snapshot. So were all three DeepSeek ids, and all nine Together ones — including the only "live" one, which carried a `Meta-` prefix Together never used, so it had never worked. OpenAI still offered `gpt-4.5-preview`, `gpt-4-32k` and `gpt-3.5-turbo-16k`, and labelled `gpt-4` as "GPT-4 Turbo". Bedrock offered a Claude 3.5 Sonnet that hit EOL in July and a Titan Text Express that is gone from the catalogue.

Re-pinning alone would have left three axes empty, so each is repopulated with what its provider serves today, preferring a rolling alias over a dated snapshot wherever one exists. On Bedrock the current Anthropic ids dropped both the date and the `-v1:0` suffix, so those are aliases too.

The list also stops trying to be a catalogue. For a provider the core-service can list live, the pinned entries only fill the screen before a key is chosen, so those axes now keep at most the current family; the axes with no listing endpoint (Vertex, Bedrock, Perplexity) stay curated, because there the pinned list is the only source. 76 entries down to 35.

No backend change is needed: the core-service does not validate the model string, and `ProviderSupportsDynamicModels` keys on provider, not model — the provider set is unchanged.

Two things a reviewer should know. Perplexity's Sonar Chat Completions is superseded by their Agent API and supported only until 2026-09-27 — the ids here are right today, but that axis needs a real integration change before then, not another re-pin. And Fireworks requires a key to list, so its three entries are the ones I could confirm against the public catalogue rather than its newest models.

On the spec: it checks shape, not freshness — nothing local can tell that a provider retired something. Run against the *old* list, only one of its six assertions fails (the cap on live-listed axes); the prefix and duplicate checks pass on both sides, so they are guards against future mistakes, not evidence for this change. `tsc -b` and `eslint` are clean and the new spec passes; the full vitest suite was still running when this was opened and I will report it here.

## Summary by Sourcery

Refresh the available model catalog so the selector offers valid current models without duplicating providers’ live listings.

Bug Fixes:
- Refresh the pinned model selections across providers to remove retired, invalid, and outdated model identifiers and replace them with currently supported models.

Enhancements:
- Reduce pinned models for providers with live model listing to the current model family while retaining curated entries for providers without listing endpoints.
- Preserve provider-specific routing and integration constraints for Bedrock and Perplexity in the curated model list.

Tests:
- Add validation for provider ownership, model prefixes, duplicate values and labels, and limits on pinned models for dynamically listed providers.

---

## Update after review

The branch moved twice since this description was written, so read the numbers here, not above.

**Retracted from the text above.** The count is **34 entries, not 35** (76 originally). The Fireworks paragraph is wrong: the three ids it defended were dead in Fireworks' JSON catalogue and were replaced. Two ids it kept were also dead — Bedrock's Llama 3.1 405B (EOL 2026-07-07) and the Fireworks trio. Two of the six original assertions used `!seen.add(x)`, which is never true, so they could not fail.

**Fixed in review (commit `33d690f`).**

- **Vertex was never swept.** It still pinned `gemini-2.5-pro/-flash/-flash-lite`, a family Google retires on **2026-10-16** (Vertex release notes, 2026-04-02). Vertex has no listing endpoint, so nothing would have covered for it. Now `gemini-3.7-flash` and `gemini-3.1-flash-lite`; Pro stays on 2.5 because 3.1 Pro is still preview on Vertex.
- **Together was emptied on a false premise.** Its public serverless catalogue does list the exact strings — the removed entry failed only for a `Meta-` prefix Together never used, which is `meta-llama/Llama-3.3-70B-Instruct-Turbo`. Three verified entries are back, so no listed provider has an empty picker when a live listing fails. A new assertion locks that in.
- **Shrinking the list silently broke `LLMConfigForm`.** It cleared a model mismatched against the API key's provider only when the model was in `availableModels`; the 45 pruned ids became "unknown", so the mismatch was saved instead of cleared. It now reads the `provider/` prefix, which both the pinned list and the live listing build.
- The cap on live-listed providers gets a named constant, and three comments that misdescribed the code are corrected.

**Verified against primary sources.** Every id was checked: OpenAI sol/terra/luna and the `gpt-5.6` alias; the OpenRouter slugs; DeepSeek V4 (`deepseek-chat`/`reasoner` really were discontinued 2026-07-24); the Fireworks slugs; Perplexity (`sonar-reasoning` removed 2025-12-15); and every Bedrock id including `us.` vs `global.` — Llama 3.1 70B genuinely has no global profile, Sonnet 4.5 genuinely has one.

**Dated follow-ups.** Perplexity Sonar until 2026-09-27; `gpt-4.1-nano` pinned in the core-service `agent_service.go:428` until 2026-10-23; Bedrock Sonnet 4.5 EOL from 2026-09-29; Vertex Gemini 2.5 Pro until 2026-10-16.


daniloleonecarneiro:
Suite result, as promised above: 10 files / 49 tests fail on this branch, and the same 10 files / 49 tests fail on a clean `origin/develop` (8e6d30d) — identical counts, so they are pre-existing and unrelated. They are ChatSidebar, ContactAvatar, tokens.contrast, Sidebar, MainLayout, MessageList, opusRecorder, NewChannel, Chat and lastSavedMark; none of them touch `availableModels`. `vitest run src/components/ai_agents/` is 12 files / 64 tests green, and `tsc -b` and `eslint` are clean.

Also amended since opening: the in-file comments were cut back to the two facts the code cannot state on its own — that this list is the pre-key view and the only source for the three providers without a listing endpoint, and Perplexity's 2026-09-27 cutoff. The rest of the reasoning lives in this description.

daniloleonecarneiro:
Follow-up after an adversarial review pass. Nine findings, and three of them were real defects in what I had opened this PR with — the branch has been amended.

**The new spec was partly fake.** Two of its six assertions used `filter(m => !seen.add(m.value))`. `Set.add` returns the Set, so the predicate is always false and those two could never fail. Proven by injecting an entry duplicating both value and label: all six passed. Rewritten with an explicit `duplicatesOf` helper, and the same injection now fails exactly those two and passes once reverted.

**Two ids I had shipped were dead.** `meta.llama3-1-405b-instruct-v1:0` is Legacy on Bedrock with EOL 2026-07-07, already past — removed. And all three Fireworks entries (`llama-v3p2-90b-vision-instruct`, `llama-v3p2-3b-instruct`, `qwen2p5-coder-7b`) are gone from its catalogue; I had kept them on the strength of a rendered docs page, and its public JSON catalogue says otherwise. Replaced with current ones.

Every remaining id has since been checked against a primary source. Confirmed correct as they stood: `gpt-5.6` really is the documented alias for `gpt-5.6-sol`; Bedrock's Sonnet 4.5, Llama 3.1 70B, DeepSeek R1, Mistral 7B and Nova Micro are all Active with the exact strings verbatim. Together is now pinned to nothing, because it publishes no catalogue without a key and an unverified id is the bug being fixed. Bedrock ids moved to `global.` where global routing exists so the picker stops pinning US residency by default. The "every provider has an entry" assertion was dropped — it is not an invariant, and it would have blocked the correct fix when Perplexity's axis has to be emptied on 2026-09-27.

76 → 31 entries. `tsc -b` and `eslint` clean, `vitest run src/components/ai_agents/` green at 12 files / 63 tests, and the 10 pre-existing suite failures are unchanged from `origin/develop`.

gomessguii (approved):

